PR #9, opened 2015-09-02
Feature/nexteuropa 5238 fix coding standards in themes ec resp 17

senseca:
Replace the other pull request closed by error (branch unfortunately deleted...)


profiles/common/themes/ec_resp_17/css/text_size_small.css
```diff
@@ -0,0 +1,4 @@
+/**
```
gervasek: "Can we remove this? Maybe not risk it in scope of this issue, there might be modules in subsites including it."

profiles/common/themes/ec_resp_17/css/print.css
```diff
@@ -6,7 +6,7 @@ body,
 input,
 textarea,
 select {
-  color: #000;
+  color: 1000;
```
gervasek: This is not valid CSS I think. Should this be #000000 or black?

profiles/common/themes/ec_resp_17/css/text_size_small.css
gervasek: @senseca  replied on JIRA task : 'Concerning "text_size_small.css", I didn't removed it in case it was used by some other sites.I don't know if it is necessary.'

OK you're right !!!

profiles/common/themes/ec_resp_17/css/print.css
senseca: Corrected!

gervasek:
Code review & Manual test OK. But we must wait the success of phpcontinuous test before merging.

PR #12, opened 2015-09-03
NEXTEUROPA-5408: Fix coding standards in communities

senseca:


profiles/common/modules/features/communities/communities.module
```diff
@@ -4,51 +4,83 @@
...
+        $output .= l($letter, $url, array(
+          'attributes' => array(
+            'type' => 'small',
+            'action_bar' => 'first',
+            'btn_group' => 'first',
+            'class' => array('hidden-phone'),
+          ),
+          )
+        );
       }
```
pfrenssen: The indentation here is not correct, you can see from lines 40+41 that the parentheses are on the same indentation level. The code sniffer doesn't seem to detect it.

These inline arrays in function calls can be quite hard to read. It is often suggested to solve it by putting the value in a separate variable. This will make it much easier to read. For example like this:

```
$options = array(
  'attributes' => array(
    ...
  ),
);
$output .= l($letter, $url, $options);
```

This same problem appears several times in this file.

profiles/common/modules/features/communities/communities.module
```diff
@@ -335,55 +355,55 @@ function _communities_get_url_in_group($path, $original_path) {
...
-/*
-* _to_arg function
-*/
-
-
-function group_name_to_arg($arg=NULL) {
+/**
+ * Implementation of _to_arg function.
+ */
+function group_name_to_arg($arg = NULL) {
```
pfrenssen: This is not a hook implementation, so this documentation is not correct. I've looked up how they do this in core and I found inspiration by looking at `user_uid_optional_to_arg()`.

```
/**
 * Returns $arg or the user ID of the current user if $arg is '%' or empty.
 */
function user_uid_optional_to_arg($arg) {
  ...
}
```

So they document what exactly is returned, but the parameter and return value are not described since it is always the same in these `_to_arg()` functions. So in this case it could be something like:

```
/**
 * Returns the sanitized group title or 'none' if there is no group context.
 */
function group_name_to_arg($arg = NULL) {
  ...
}
```

profiles/common/modules/features/communities/communities.module
```diff
@@ -335,55 +355,55 @@ function _communities_get_url_in_group($path, $original_path) {
...
 
 
-/*
-* _load function
-*/
+/**
+ * Implementation of _load function.
+ */
 function group_name_load($arg) {
   global $language;
```
pfrenssen: A `_load()` function should be fully documented. See for example `node_load()` and `user_load()`. This function has a very unfortunate name, since it does not actually load a "group name", instead it loads the group _node_ but then only returns the node ID instead of the full node :-/

Anyway let's just focus on the coding standards.

For example:

```
/**
 * Returns the ID of the Community node corresponding to the given group name.
 *
 * @param string $arg
 *   The group name, or 'none' if no ID should be returned.
 *
 * @return string | 0
 *   The node ID of the Community node, or the integer 0 if no group name was given.
 */
```

pfrenssen:
Thanks! Merging into develop.

PR #23, opened 2015-09-07
NEXTEUROPA-5472: fix feature set for webtools;fix feature set (check

gervasek:
- fix feature set for webtools
- fix feature set (checkbefore enabling/disabling)


senseca:
I always have the message : 
"Notice: Undefined variable: check_node in feature_set_disable_feature_set() (line 183 of /ec/dev/server/fpfis/webroot/sources/senseca/platform-test/profiles/multisite_drupal_standard/modules/contrib/feature_set/feature_set.module)."
It appears (in watchdog) when I disable the WebTool feature.

PR #27, opened 2015-09-09
MULTISITE-7137

PetricaBaragan:
Call to legacy function curl_http_request. Please use chr_curl_http_request instead.


senseca:
I have this message during the build : 
" Unable to patch chr with chr-deprecated_call-2512054-2.patch. "
Indeed, the patch is already applied "patches/chr-deprecated_call-5588.patch"...

PetricaBaragan:
Indeed, the patch is already applied "patches/chr-deprecated_call-5588.patch"

PR #46, opened 2015-09-16
NEXTEUROPA-5559: Use the generic, contributed git pre-push hook.

pfrenssen:
This contains some improvements and bugfixes that were made for the Subsite
Starterkit but will also benefit the platform.


resources/scripts/composer/post-install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# Symlink the git pre-push hook to its destination.
+if [ ! -h ".git/hooks/pre-push" ] ; then
+  ln -s "../../vendor/pfrenssen/phpcs-pre-push/pre-push" ".git/hooks/pre-push"
```
senseca: Just one remark about this : 
If the symbolic link is already existing (for example : linked to resources/git-hooks/pre-push, which is deleted), the symlink is not recreate and doesn't work...
Maybe is it possible to check if it is linked to "../../vendor/pfrenssen/phpcs-pre-push/pre-push" ?

pfrenssen: Good suggestion. I solved it by removing the symlink if it is present, and then creating the new symlink.

PR #50, opened 2015-09-18
Nexteuropa newsroom feature for multisite platform - MULTISITE-7232

LOBsTerr:


jazio:
@LOBsTerr Please redo this pull request as it is supposed to be done not in this folder but in profiles/common/modules/features

PR #51, opened 2015-09-18
Hotfix/nexteuropa 5486-splash screen

gervasek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-5486


pfrenssen:
* Checks failed but this is OK since it is a hotfix.
* The variable `splash_screen_home_page` is removed, but it is still referenced in two places: in the `splash_screen.info` file and in `splash_screen_preprocess_splash()`.
* The settings form at path `admin/config/splash_screen/settings` is being removed from `hook_menu()`, but the code still remains in `includes/splash_screen.admin.inc`. I could not find any more references to this form, I guess it can be removed?

sites/all/modules/features/custom/splash_screen/splash_screen.module
```diff
@@ -35,26 +22,18 @@ function splash_screen_menu() {
...
-    }
-    return FALSE;
+function splash_screen_boot() {
+  // If empty, $_GET['q'] is rewritten by splash screen path.
+  if (empty($_GET['q'])) {
+    $_GET['q'] = SPLASH_SCREEN_PATH;
   }
 }
 
+
```
pfrenssen: Coding standards: a stray empty line is added here.

sites/all/modules/features/custom/splash_screen/splash_screen.module
```diff
@@ -35,26 +22,18 @@ function splash_screen_menu() {
 }
 
 /**
- * Custom callback function to check access.
+ * Implements hook_boot().
+ *
+ * Display splash page when go at /
```
pfrenssen: Coding standards: comments should end in a period.

gervasek:
- ``splash_screen_home_page`` removed from .info and .module fiels
- menu router for admin page has been restored: still used for the ``$languages_blacklist`` variable
- coding standard fixed

kalinchernev:
Hi, I quickly tested after latest changes and there are issues.
- config settings are no longer accessible, but needed. 
- page reloads but do not lead to anywhere: In our case, home page node path is "index", splash tries to reach index_en for example. Clicking on any option
- $base_url at splash_screen_preprocess_splash() is never used. Could be removed
- splash_screen_update_7201 seems to force site_frontpage to be same as splash page? Being unsure I can just share that we need a separation as our <front> should not be 'splash'
- features[variable][] = splash_screen_home_page at info file is not needed any more in this scenario
- _splash_screen_settings_form() settings are needed, please provide hook_menu for it

gervasek:
Hi, I quickly tested after latest changes and there are issues.
- config settings are no longer accessible, but needed. 
=> DONE
- page reloads but do not lead to anywhere: In our case, home page node path is "index", splash tries to reach index_en for example. Clicking on any option
=> I don't understand this remark, could you develop please
- $base_url at splash_screen_preprocess_splash() is never used. Could be removed
=> DONE
- splash_screen_update_7201 seems to force site_frontpage to be same as splash page? Being unsure I can just share that we need a separation as our <front> should not be 'splash'
=> no, site frontpage is restored with the splash_screen_home_page (= the previous frontpage). If the frontpage has been changed in the meantime (!= splash) we don't restore the frontpage
- features[variable][] = splash_screen_home_page at info file is not needed any more in this scenario
=> DONE
- _splash_screen_settings_form() settings are needed, please provide hook_menu for it
=> DONE

kalinchernev:
@gervasek thanks for the quick reaction. Running the update hook, functionally, transition between url '/' (splash) and '/index' works well.

However, even though variable_set is not our 'index', going to the home page, Drupal still does not accept the page as <front>. So, context settings will not work, as well as any CSS rules targeting .front class.

Strangely, now we have a situation in which neither / is the homepage, nor /index, nor index_en. 

The scope of the ticket is to remove the redirect which is ok now, however I don't feel comfortable merging without having confirmation from my team that we are ok to not have the <front> in settings and .front in body tag. I'll try to find this information asap

kalinchernev:
Merging the fix as it solves the scope of NEXTEUROPA-5486. Will open a follow up ticket for the front page negotiation behavior.

PR #57, opened 2015-09-21
NEXTEUROPA-5707: Adding a check for existing 2nd language argument

kalinchernev:
When we have 2nd-language GET argument, language_selector_page should respect.

This is when it's not respected:
not_respected

And this is after it's respected:
respected


gervasek:
Close, see https://github.com/ec-europa/platform-dev/pull/66

kalinchernev:
Ok, will test the other one

PR #58, opened 2015-09-22
NEXTEUROPA-5619 removed javascript alterations to php. Moved language…

haringsrob:
Hi,

This pull requests solves an issue where every link becomes absolute as soon as the 2nd-language parameter is active.

1. The 2nd-language parameter is part of the language_selector_page module so I moved the hooks into the correct folder.
2. The alterations in the urls were done by Javascript. This would result in a broken website if the enduser does not have javascript enabled. So I have removed the javascript file and implemented this in a more Drupal way.


haringsrob:
Closing for now, need additional modifications,

PR #60, opened 2015-09-23
NEXTEUROPA-5619 removed js, placed hooks in correct module

haringsrob:
I removed the javascript that alters the url and replaced it with drupal hooks.

Also there were some parts related to the page level switcher in the nexteuropa multilingual module. So I have moved those.


sites/all/modules/custom/language_selector_page/language_selector_page.module
```diff
@@ -121,10 +121,8 @@ function language_selector_page_preprocess_block(&$variables) {
     if (!empty($variables['elements']['other']['#markup'])) {
       foreach ($variables['elements']['other']['#markup'] as $code => $lang) {
         $options = array(
-          'language' => $lang,
```
gervasek: Why you remove this parameter ?

haringsrob: Hello,

It has been removed because this will also change the interface language.

So when the page is: 
*page_en*
and you click in the page level switcher on *dutch* the expected value is:
*page_en?2nd_language=nl*
but instead, if you have this line the result is:
*page_nl?2nd_language=nl*

The page level switcher should leave the interface untouched

haringsrob:
Updated it with a fix for the autocomplete.

sites/all/modules/custom/language_selector_page/language_selector_page.module
```diff
@@ -142,3 +152,88 @@ function language_selector_page_preprocess_block(&$variables) {
...
+/**
+ * Implements hook_language_fallback_candidates_alter().
+ *
+ * Change the language fallback order.
+ */
+function language_selector_page_language_fallback_candidates_alter(array &$fallback_candidates) {
+  // First we get our url parameters.
+  $parameters = drupal_get_query_parameters();
+
+  // Add the default language as first fallback.
```
gervasek: This code should be on nexteuropa_multilingual feature, like on the commit d2968dc.
You can keep the second language in this module if you want, but the code for the default language must be on nexteuropa_multilingual feature.

gervasek:
Approved

PR #92, opened 2015-10-06
feature/NEXTEUROPA-5974 check chr patches

champcy:


resources/patches/chr-jira-update-custom-fields-using-rest-php-and-curl.patch
```diff
@@ -0,0 +1,16 @@
+--- chr.module	2014-10-17 14:29:42.338714671 +0200
++++ chr.module	2014-10-17 14:33:54.994184777 +0200
+@@ -448,6 +456,13 @@
+       curl_setopt($ch, CURLOPT_HTTPGET, TRUE);
+       $valid_method = TRUE;
+     break;
++    // FPFIS fix jira REST remoting
++    // https://answers.atlassian.com/questions/72475/jira-update-custom-fields-using-rest-php-and-curl
++    case 'PUT_JIRA':
```
pfrenssen: This is not a good solution, there is no HTTP request type called "PUT_JIRA".

PR #97, opened 2015-10-07
MULTISITE-5739 Translate Overview Removal

jazio:


jazio:
@gervasek Could you please review this request?

PR #104, opened 2015-10-09
NEXTEUROPA-5948: revert part of commit 59630b1

gervasek:


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.install
```diff
@@ -39,6 +39,10 @@ function nexteuropa_multilingual_install() {
   // Set translatable states for TMGMT Workbench module.
   multisite_config_service('system')->setVariable('tmgmt_workbench_moderation_states', array('validated', 'published'));
 
+  // Disable multisite_drupal_language_negociation (redundant with
+  // nexteuropa_multilingual).
+  module_disable(array('multisite_drupal_language_negociation'));
```
pfrenssen: This is not a good solution. This will silently disable `multisite_drupal_language_negociation` when `nexteuropa_multilingual` is installed. This would be unexpected for a site owner.

There are different solutions for this. I will open a discussion in issue NEXTEUROPA-5948 to agree on the best solution.

gervasek: This line was added 6 months ago on the platfrome, and I deleted it by error on the previous commit.This code revert the mistake.

PR #106, opened 2015-10-12
Master

PetricaBaragan:
Remove the deprecated dependencies as per MULTISITE-7395


PetricaBaragan:
I will do a separate pull request for develop.

PR #133, opened 2015-10-22
NEXTEUROPA-6259 : Updated patch

senseca:


profiles/common/modules/features/cce_basic_config/cce_basic_config.install
```diff
@@ -1643,3 +1643,22 @@ function cce_basic_config_update_7208() {
...
+ * Drupal 7.41 update.
+ *
+ * Added here in case the update 7080 of system module was already
+ * executed on existing sites.
+ * Convert the 'format' column in {date_format_locale}
+ * to case sensitive varchar.
+ */
+function system_update_7209() {
+  $spec = array(
+    'description' => 'The date format string.',
```
pfrenssen: This function has moved from the `system` module to the `cce_basic_config` module, so it should be renamed from `system_update_XYZ()` to `cce_basic_config_XYZ()`.

Can you also add this line to the documentation? This is the standard way of referring to other functions.

```
/**
 * ...
 * @see system_update_7080()
 */
```

PR #135, opened 2015-10-23
Fixed module not taking into account both files and file contents

imanoleguskiza:


degliwe:
fixed from 2.0.*  (master was already 2.1.*) see my comment here. https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-6339

PR #151, opened 2015-10-28
MULTISITE-5657 Temporary solution to solve the issue.

tatarbj:
It should solve the problem what is reported by rawmaterials, but this solution cannot be a long term one! I've already start to explain it in a new ticket and take a look in all of the platform features for the same pattern which should be obviously forbidden to follow!


tatarbj:
Can someone link a documentation about the current workflow, if i want to fix something on the platform level (like i did with this issue) which branch should i use as a basic one?! Very very annoying... I'm starting to do the same change from master.

PR #154, opened 2015-10-29
Reworked tmgmt_poetry module

imanoleguskiza:
Reworked tmgmt_poetry module to follow latest requirements from DGT


degliwe:
use PR 202 instead of this one (154)
https://github.com/ec-europa/platform-dev/pull/203

PR #167, opened 2015-11-03
NEXTEUROPA-5585 : language suffix on splash screen

gervasek:
Use the language option to retrieve the link (instead of html attribute)
language

https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-5585


degliwe:
See justification on the ticket

PR #173, opened 2015-11-04
NEXTEUROPA-6826: [Poetry] Content alias is lost after completing a job

ademarco:


degliwe:
Branch rebased on 2.0.64 so this PR is not relevant anymore

PR #183, opened 2015-11-05
fix my own mistake

degliwe:
I poorly resolved a conflict and, mea culpa.


degliwe:
@senseca @gervasek please double check and merge. this time it should be ok.

PR #197, opened 2015-11-13
hotfix/MULTISITE-7703: Change the fancybox loading location

PetricaBaragan:


mathiasselleslach:
Dear @PetricaBaragan,

the right fix would be to fix the symlinks.
because only the communities profile has the wrong path (/source/), standard does NOT have this.
So this fix, will fix communities but will break standard.

I quickly checked this with @degliwe, he agreed to fix the symlinks in MULTISITE-7717

I will take care of linking all of the issues together.

degliwe:
fixed in https://github.com/ec-europa/platform-dev/pull/199

PR #201, opened 2015-11-13
NEXTEUROPA-6099 - Language switcher position fixed

yhuard:


gervasek:
Close, a new one will be created with the fix of block region.

PR #203, opened 2015-11-16
Hotfix/nexteuropa 2478

degliwe:
an attempt to move the relevant commits from develop to master


profiles/common/modules/custom/tmgmt_poetry/controller/tmgmt_poetry.controller.job.inc
```diff
@@ -0,0 +1,91 @@
...
+          _tmgmt_poetry_insert_item_status($job_item->tjiid, $job->target_language, "Pending", date('d/m/Y', strtotime($origin_job->settings['delai'])));
+        }
+      }
+    }
+
+    // IF A POETRY TRANSLATOR HAS ACCEPTED
+    // A TRANSLATION REQUEST WE SHOULD UPDATE
+    // THE RELATED ADDITIONAL SUB TRANSLATION JOBS
+    // WITH A REFERENCE RETURNED BY POETRY.
+    if ($origin_job->state == TMGMT_JOB_STATE_ACTIVE && (strpos($origin_job->reference, 'MAIN_' . $origin_job->tjid . '_POETRY') !== FALSE)) {
```
pfrenssen: I LIKE THAT THESE COMMENTS
READ LIKE POETRY BUT THEY SHOULD BE
IN HAIKU FORMAT

profiles/common/modules/custom/tmgmt_poetry/inc/tmgmt_poetry.webservice.inc
```diff
@@ -0,0 +1,470 @@
...
+  $items['tmgmt_poetry/service_callback'] = array(
+    'page callback' => '_tmgmt_poetry_service_callback',
+    'access callback' => TRUE,
+  );
+  $items[drupal_get_path("module", "tmgmt_poetry") . "/wsdl/PoetryIntegration.wsdl"] = array(
+    'page callback' => '_tmgmt_poetry_wsdl',
+    'access callback' => TRUE,
+  );
+  return $items;
+}
```
pfrenssen: Just a note: main functions and hook implementations should always be put in the module file. They should not be put in include files. Include files should contain secondary code that can be included at runtime.

The only way that this will work is if the include file is hard included in the module file using `require_once()` which defeats the purpose of using include files using `module_load_include()`.

pfrenssen:
A bunch of files in `tmgmt_dgt_html` have been made executable - their permissions have changed from 644 to 755. This should be fixed, it's not a good practice to have PHP files executable.

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -1,300 +1,57 @@
...
- */
-function tmgmt_poetry_menu() {
-  $items = array();
-  $items['tmgmt_poetry/service_callback'] = array(
-    'page callback' => '_tmgmt_poetry_service_callback',
-    'access callback' => TRUE,
-  );
-  return $items;
+  module_load_include('inc', 'tmgmt_poetry', 'inc/tmgmt_poetry.webservice');
 }
```
pfrenssen: This is pointless. Don't use `hook_init()` to include other files. This will be executed on every request.

profiles/common/modules/custom/tmgmt_poetry/inc/tmgmt_poetry.access_control.inc
```diff
@@ -0,0 +1,94 @@
...
+    $query->add_where(isset($query->options['group']),
+      db_or()
+        ->condition(
+          db_or()
+            ->condition('reference', 'SUB_%', 'NOT LIKE')
+            ->isNull('reference')
+        )
+    );
+  }
+}
```
pfrenssen: This is a hook implementation which should be moved into the main module file. The same goes for all other hook implementations in this file.

profiles/common/modules/custom/tmgmt_poetry/inc/tmgmt_poetry.webservice.inc
```diff
@@ -0,0 +1,470 @@
...
+  return FALSE;
+}
+
+/**
+ * Function available to call from our webservice.
+ */
+function FPFISPoetryIntegrationRequest($user, $password, $msg) {
+
+  watchdog("POETRY", "Poetry accessed callback");
+  watchdog('tmgmt_poetry: translation response', $msg, array(), WATCHDOG_DEBUG);
```
pfrenssen: This seems to be debugging info. Is it really necessary to log this every time this is called?

profiles/common/modules/custom/tmgmt_poetry/inc/tmgmt_poetry.webservice.inc
```diff
@@ -0,0 +1,470 @@
...
+ */
+function _tmgmt_poetry_authenticate($user, $password) {
+  // Get Poetry configuration.
+  $poetry = variable_get("poetry_service");
+  if (($poetry['callback_user'] == $user) &&
+      ($poetry['callback_password'] == $password)) {
+    return TRUE;
+  }
+  return FALSE;
+}
```
pfrenssen: This is overly verbose. This is equivalent to the following:

```
return $poetry['callback_user'] == $user && $poetry['callback_password'] == $password;
```

These parameters are not documented but the look like strings. If they are strings then you should always do a literal comparison using `===` instead of using `==` which allows casting. The input types are not checked, so using the latter will have unintended consequences in case other types are passed. For example if `$user` equals `TRUE` then this will return `TRUE`.

pfrenssen:
This looks good overall. I haven't looked at the code in detail but I made some remarks of things that popped out to me. I trust this has been reviewed properly by one of our colleagues?

The only thing I would like to see fixed is the incorrect permissions. Some PHP files have been changed to 775, please change them back to 664 before merging this in.

senseca:
I corrected : 
* comments
* permissions
* include in hook_init
I didn't moved hooks included in .inc files. It could be done in a follow up task.
If it is OK, it can be merged?

PR #207, opened 2015-11-17
CPHP now supports release as well

gboddin:
CPHP now supports release as well, removing .travis to avoid double uploads


gboddin:
Should go to master

PR #210, opened 2015-11-17
NEXTEUROPA-6826 : Alias lost after job completing

senseca:


senseca:
Closed : not to be merged into develop but into master!

PR #224, opened 2015-11-19
NEXTEUROPA-5585: Splash screen: Link to language versions does not work if language uses a custom prefix

pfrenssen:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-5585

### DIFF
--- a/tests/features/splash_screen.feature
+++ b/tests/features/splash_screen.feature
@@ -49,9 +49,6 @@ Feature: Splash Screen features
     And I should see the link "Deutsch"
     And I should not see "Български"
     And I should not see "Français"
-    # Clean up after the fact.
-    # @todo Remove this once NEXTEUROPA-5519 is in.
-    And I run drush "vdel" "splash_screen_blacklist --yes"
 
   Scenario: Being able to change the splash screen title
     Given I am logged in as a user with the 'administrator' role
--- a/tests/features/splash_screen.feature
+++ b/tests/features/splash_screen.feature
@@ -26,16 +26,18 @@ Feature: Splash Screen features
     And I should see the link "Български"
 
   @api
-  Scenario: Links on splash screen pages are correct
+  # Regression test for a bug that broke the Portuguese (pt-pt) link.
+  # See https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-5585
+  Scenario: Test language with a custom prefix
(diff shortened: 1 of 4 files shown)


tests/features/splash_screen.feature
```diff
@@ -50,3 +49,6 @@ Feature: Splash Screen features
     And I should see the link "Deutsch"
     And I should not see "Български"
     And I should not see "Français"
+    # Clean up after the fact.
+    # @todo Remove this once NEXTEUROPA-5519 is in.
+    And I run drush "vdel" "splash_screen_blacklist --yes"
```
gervasek: Can be deleted (not necessary to wait the fix of NEXTEUROPA-5519)
Currently, enabled module(s) are uninstalled at the end of scenario   (see ``cleanModule()``), and the variable is deleted on the ``splash_screen_uninstall`` hook.
(I did not realize that the background is executed for each scenario (so the desinstall of modules in the @afterscenario is executed for each scenario)

tests/features/splash_screen.feature
```diff
@@ -25,19 +24,19 @@ Feature: Splash Screen features
...
 
-  @api
-  Scenario: Links on splash screen pages are correct
+  # Regression test for a bug that broke the Portuguese (pt-pt) link.
+  # See https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-5585
+  Scenario: Test language with a custom prefix
     Given I am logged in as a user with the 'administrator' role
-    When I go to "admin/config/regional/language/edit/fr"
-    And I fill in "edit-prefix" with "fr-prefix"
+    And the "pt-pt" language is available
```
gervasek: Why we don't use the existing function  {{these following languages are available}} ?

pfrenssen: Because it was not understood by the original reporter that this test covered their use case, since it was not using the language they reported. For me this change is not important at all, but I think people familiar with the issue might search for "pt-pt" in the codebase.

tests/features/splash_screen.feature
pfrenssen: Thanks, I didn't realize they were uninstalled, I thought they were merely disabled. Will remove this.

tests/features/splash_screen.feature
gervasek: You're right , it's a good idea.

PR #229, opened 2015-11-20
NEXTEUROPA-7077 : Behat tests for alias lost

senseca:


gervasek:
Close, see https://github.com/ec-europa/platform-dev/pull/245

PR #231, opened 2015-11-20
hotfix/MULTISITE-7754

mnfriend:
https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-7754


senseca:
Useless. The feature newsletters_og is unused. It was removed.

PR #259, opened 2015-11-30
NEXTEUROPA-5995: delete core profile

gervasek:
move core profile installation into cce_basic_config


profiles/multisite_drupal_communities/multisite_drupal_communities.install
```diff
@@ -10,8 +10,7 @@
  * Perform actions to set up the site for this profile.
  */
 function multisite_drupal_communities_install() {
-  include_once DRUPAL_ROOT . '/profiles/multisite_drupal_core/multisite_drupal_core.install';
-  multisite_drupal_core_install();
+  cce_basic_config_post_install();
```
gervasek: It's necessary to run ``cce_basic_config_post_install`` after the installation of all multisite modules/features, that's why we cannot run ``cce_basic_config_post_install()`` at the end of ``cce_basic_config_install``

PR #264, opened 2015-11-30
NEXTEUROPA-7230 fix multisite_rules_configurations overriden

champcy:


profiles/common/modules/features/nexteuropa_editorial/nexteuropa_editorial.install
```diff
@@ -91,7 +91,6 @@ function nexteuropa_editorial_install() {
   multisite_config_service('title')->replaceTitleField('node', 'editorial_team', 'title');
 
   // Add NextEuropa specific moderation states.
-  multisite_config_service('workbench_moderation')->createModerationState('validated', 'Validated', 'Ready to be reviewed', 0);
```
senseca: If we removed this line, should we not add a dependency to multisite_rules_configurations?

senseca:
MULTISITE_DRUPAL_TOOLBOX_FEATURES_REVERT_BLACKLISTED have to be updated.

profiles/common/modules/features/nexteuropa_editorial/nexteuropa_editorial.install
champcy: Yes true, but we intend to remove the module multisite_rules_configuration which is quite useless.

PR #288, opened 2015-12-10
MULTISITE-7962 patch for fullcalender module

mnfriend:
https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-7962


mnfriend:
Hi @degliwe 
I actually didn't try to apply the patch so it might break the module. So please don't merge until I confirm that the patch doesn't break the module.

Best 
Mohammed

Fefaine:
Ticket is closed.

PR #301, opened 2015-12-16
MULTISITE-8047 fix xss filter for tokens used in a tag as href

mnfriend:
https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-8047


janoka:
Tested on a brand new Platform install, perfectly work the token for me.

PR #303, opened 2015-12-22
MULTISITE-8159 translate properly values of select list

champcy:


senseca:
Corrected in another task.

PR #307, opened 2016-01-05
MULTISITE-7359 Added wysiwyg patch for object vs array features bug

mathiasselleslach:


tatarbj:
I've tested it and it seems to be OK, the PR is accepted by me.
Btw we should post it on d.org as well.

mathiasselleslach:
Good idea @tatarbj !
I will submit it to d.org ;-)
Thanks for the review

PR #330, opened 2016-01-18
MULTISITE-8308: preventing trigger of a second batch process.

SeyVaneerdewegh:
See https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-8308 for more info.


netlooker:
I've checked proposed solution for hook_update_N() implementation for following cases:
1. Batch API implementation inside hook_update - there weren't any errors.
2. Custom operations against taxonomy_term_delete - no problems found.

I attached .install file with some custom code to test both scenarios as a part of https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-9075

PR #331, opened 2016-01-19
NEXTEUROPA-8276 translate taxonomy as list field results in error

champcy:


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -343,7 +343,7 @@ function nexteuropa_multilingual_preprocess_field(&$variables) {
       $variables['label'] = i18n_field_translate_property($instance, 'label', $lang);
     }
     // Translate field value (for select list field).
-    if ($translated_value = i18n_field_translate_allowed_values($field, $lang)) {
+    if ($field['type'] !== 'taxonomy_term_reference' && $translated_value = i18n_field_translate_allowed_values($field, $lang)) {
```
gervasek: I think it would be better to limit field type to "select list" (instead of exclude taxonomy term):
```if ($field['type'] == 'list_text' && .....)```

kalinchernev:
Personally I agree to rather limit to select list with a single check as @gervasek suggests, @haringsrob should also give an opinin.

haringsrob:
Its good if field labels are still following content language. (which seems the case).

They way we handle the list doesn't really matter, we should just never check for "referenced" entity's of any type.

PR #332, opened 2016-01-21
NEXTEUROPA-5609 patch for entityreference to allow altering results i…

haringsrob:
For one of our new implementation we are extending the entity reference form widget.

For an optimal experience (without writing duplicate code) we would like to hook into the display widget of entity reference.

As this is a global improvement for the entity reference field, we have created a feature request with the implementation to the entity reference issue queue.

Unfortunately  after a few attempts (irc/twitter) I could not get in contact with the maintainers to speed things up.

link: https://www.drupal.org/node/2651982


degliwe:
This PR is from a feature branch to the develop branch.
I assume you want this on prod asap...so a hotfix to the master seems more appropriate.
Can you please confirm?

PR #343, opened 2016-01-29
MULTISITE-7598

jazio:
I have removed the references to multisite_crop_and_resize but not the feature/module physically.

https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-7061

### DIFF
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -349,9 +349,6 @@ projects[media][patch][] = https://www.drupal.org/files/issues/media-wysiwyg-ove
 ; https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-3650
 projects[media][patch][] = http://www.drupal.org/files/issues/file_entity-remove-contextual-links-2401811-11.patch
 
-projects[media_crop][subdir] = "contrib"
-projects[media_crop][version] = "1.4"
-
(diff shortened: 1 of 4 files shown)


resources/multisite_drupal_standard.make
```diff
@@ -349,9 +349,6 @@ projects[media][patch][] = https://www.drupal.org/files/issues/media-wysiwyg-ove
 ; https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-3650
 projects[media][patch][] = http://www.drupal.org/files/issues/file_entity-remove-contextual-links-2401811-11.patch
 
-projects[media_crop][subdir] = "contrib"
```
gervasek: if ``media_crop`` is removed from the platfrome, the uninstall of this module cannot be done

jazio: @gervasek Thanks for reviewing it. I think I miss a point. When is the make file being triggered? According to my understanding the make file is being triggered only for new websites. Could you help me understand more what to do next or try further handling yourself in your best way?

Same with Language Tools features: https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-5739, a redundant feature that we agreed with Wesley it should be removed. It also creates a problem when activated by overriding the whole content translation UI with its own.

PR #346, opened 2016-02-02
NEXTEUROPA-8612: fix field group version on makefile

gervasek:


netlooker:
I have checked and version 1.5 for field_group module is already included in .make files in the resources folder for develop and master. File drupal-org.make was moved in this commit d301a082179f5ca70978c6e50 . I'm assuming that this pull request is related to older platform version which is still used for some of subsites.

PR #351, opened 2016-02-04
NEXTEUROPA-8555: TMGMT does not process links

gervasek:
clone of https://github.com/claudiu-cristea/platform-dev/commit/6d857c53c19325af28e62dd3e3fdb5e05407c88f to run CPHP.


resources/multisite_drupal_standard.make
```diff
@@ -524,6 +524,7 @@ projects[tmgmt][download][branch] = 7.x-1.x
 projects[tmgmt][download][revision] = bd307cb0cdf55a20092f616aeb0c39bd918aef5d
 projects[tmgmt][download][type] = git
 projects[tmgmt][subdir] = contrib
+projects[tmgmt][patch][] = patches/support_for_link_field-2489134-8.patch
```
gervasek: For contributed patches (not customised by us) it is better to use the patch of drupal.org:
``projects[tmgmt][patch][] = https://www.drupal.org/files/issues/support_for_link_field-2489134-9.patch``

PR #362, opened 2016-02-23
Feature/nexteuropa 8981

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-8981


profiles/common/modules/custom/workbench_moderation_cron/workbench_moderation_cron.module
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+/**
+ * Implements hook_cron() for workbench_moderation_cron_cron().
+ */
+function workbench_moderation_cron_cron() {
+  global $user;
+
+  $role = user_role_load_by_name("editor");
+  $user->roles = $user->roles + array($role->rid => $role->name);
```
gervasek: Add a role to the global user seem to be risked... (if the user object is saved on other hook_cron, the role will be keept)

gervasek: Can return an error if the "editor" role doesn't exist.

PR #385, opened 2016-03-04
NEXTEUROPA-7106: Drupal templates, modules and features

ademarco:


yhuard:
@ademarco thanks for your feedbacks. Actually `ec_resp` doesn't depend on `Drupal Bootstrap` (aka `bootstrap`). Currently, `next_europa` theme depends on `bootstrap` but I removed it with @degliwe's approval so I think it's pretty safe to update `Drupal Bootstrap` as long as the new theme supports it well.

Regarding the symlinks, I didn't add them to the communities profile because this profile is meant to be removed in a not-so-far future, in favor of standard profile.

ademarco:
Ok great then it's all good I'd say, please go ahead and merge into develop.

PR #389, opened 2016-03-07
Feature/nexteuropa 6330 remote browser

champcy:


profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
```diff
@@ -167,11 +167,14 @@ function nexteuropa_token_ckeditor_pre_render_text_format($element) {
 /**
  * Implements hook_custom_theme().
  */
+
+/*
```
netlooker: Is this function going to be used in future? @champcy

PR #392, opened 2016-03-14
behat test for tmgmt poetry for pt-pt

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-8342


ademarco:
Closed in favour of #403

PR #393, opened 2016-03-14
NEXTEUROPA-9688 : hotfix js_injector - wrong URLs for JS Injector JS files

nic0pad:
For the MULITSITE-9219 ticket


degliwe:
wrong target

PR #404, opened 2016-03-18
MULTISITE-8903: fix display of username

gervasek:


degliwe:
404... can't find the PR.

PR #408, opened 2016-03-23
Hotfix/nexteuropa 9874

degliwe:


gboddin:
This patch didn't fix the issue :(

PR #414, opened 2016-03-25
HOTFIX/NEXTEUROPA-9874 Reopening hotfix, previous doesn't work

gboddin:
Here we make sure locale.inc will never load any remote file trough file_get_contents
and dispatch an event to the watchdog for further debug

Previous fix didn't work, this one will act as a final barrier against bad coding.

See https://www.drupal.org/node/2385069 for the ongoing discussion

Also all found affected modules have been fixed : 

- wysiwyg
- multisite_maps (frontend???)
- multisite_twitterblock (frontend !!!)


gboddin:
cc @tatarbj

gboddin:
Previous fix actually worked ( partly since multiple modules are part of the issue ).

2.1.97 did revert 2.1.96 release.

Released 2.1.98 from master again @voidtek

resources/patches/core-locale-prevent-remote-loading.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/multisiteplayground_master_production.2.1/includes/locale.inc b/multisiteplayground_master_production.2.1/includes/locale.inc
+index 82c55e5..e086f0d 100644
+--- includes/locale.inc
++++ includes/locale.inc
+@@ -1485,6 +1485,10 @@ function _locale_parse_js_file($filepath) {
+   $parsed_url = drupal_parse_url($filepath);
+   $filepath = $parsed_url['path'];
+   // Load the JavaScript file.
```
gboddin: Here we make sure file_get_contents never gets remote URI as it would bypass the drupal_http_request method.

This happens when someone uses remote uri with drupal_add_js but forgets to set the type as external.

The locale module then tries to get its content anyway.

resources/patches/wysiwyg-js-url-9874.patch
```diff
@@ -2,22 +2,12 @@
 # See NEXTEUROPA-9874 
 --- wysiwyg.module
 +++ wysiwyg.module
```
gboddin: Let's avoid changing the behavior :

resources/patches/wysiwyg-js-url-9874.patch
```diff
@@ -2,22 +2,12 @@
...
-+          $files_path = $wrapper->getDirectoryPath() . "/" . file_uri_target($uri);
-+
-           $init_exists = file_exists($uri);
-           if (!$init_exists) {
-             $js_path = dirname($uri);
 @@ -348,7 +351,7 @@
              drupal_add_js($init, array('type' => 'inline') + $default_library_options);
            }
            else {
 -            drupal_add_js(file_create_url($uri), $default_library_options);
```
gboddin: And just force it external as it should be :

PR #417, opened 2016-03-29
NEXTEUROPA-8807 : install and enable the fast404 module

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-8807


gboddin:
@degliwe  Looks fine to me , it bypasses a lot of checks/logs for missing css/js resources (like loading analytics stuff) while keeping good behavior on normal pages

PR #427, opened 2016-04-01
Fix all perf issues with some magic sauce

gboddin:
This will fix a lot of issues with drupal with some magic sauce.

Thanks to Douglas Adams


tatarbj:
Are you really sure we should learn a framework if we have drupal? ;)

gboddin:
Well that would fix the perf issues ;)

gboddin:
It just won't fix Jim ...

PR #452, opened 2016-04-13
nexteuropa-9887: Behat's xss test of the caption field of an image

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-9887


voidtek:
replaced by the PR https://github.com/ec-europa/platform-dev/pull/453

PR #463, opened 2016-04-15
NEXTEUROPA-9797: Provide update hook to enable nexteuropa_laco

ademarco:


jfhovinne:
Module is correctly enabled during update so merging #463.

PR #485, opened 2016-04-25
MULTISITE-9253 Add symlink for nexteuropa_newsroom feature.

tatarbj:


tatarbj:
it's just ridiculous i've solved all of the issues what were reported by cphp then now there is something new with phantomjs, is it a f*cking neverending story or what?!?!?!

PR #528, opened 2016-05-10
MULTISITE-7126: Remove the dependencies of field, migrate, bean_admin,…

degliwe:
… views, context, print, subscriptions and ds ui modules.


Fefaine:
replaced by https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-597

PR #536, opened 2016-05-10
Feature/nexteuropa 7552

degliwe:


tests/features/bootstrap/FeatureContext.php
```diff
@@ -380,6 +380,52 @@ protected function transformRowTable(TableNode $table_node, array $aliases) {
...
 
     return new TableNode($table);
+}
+
+  /**
+   * Prepare for PHP errors log.
+   *
+   * @BeforeStep
+   */
+  public static function preparePhpErrors($event) {
```
degliwe: $event is not necessary

PR #538, opened 2016-05-10
Multisite 8229

degliwe:


Fefaine:
Replaced with https://github.com/ec-europa/platform-dev/pull/2267

PR #561, opened 2016-05-13
NEXTEUROPA-10830: update entity_translation to 1.0-beta5

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-10830


Fefaine:
Rebased to support 2.1

PR #579, opened 2016-05-20
Theme/nexteuropa 7528

fab01:
Added new button "HOME" on the responsive navigation, replaced old menu button (arrow down) with "MENU" button, hidden "Burger Button" designed to open the sidebar according to the IPG compliant Responsive behavior.


degliwe:
Please do not add 3rd party code to the repo, instead you should use the  make file https://github.com/ec-europa/platform-dev/blob/master/resources/multisite_drupal_standard.make#L700

PR #619, opened 2016-06-02
Add a new test for german body

Fefaine:


Fefaine:
Test already exists but only in master. Close and create a new PR to deploy test to 2.1

PR #626, opened 2016-06-02
MULTISITE-10728 move patch from DAE subsite to MS. Fix DB error on fe…

batigolix:
…eds module install


Fefaine:
We created this ticket instead.
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-11322
This patch is included in the latest version

PR #627, opened 2016-06-02
Update module version in makefile

drupol:
- Removed outdated patches (only those that has been included in the next new version)


Fefaine:
Hi Pol :)
For module upgrades we create 1 ticket per upgrade and one pull request per ticket.
For example for "chosen"  we have this ticket https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-11183
We do like this because we need to test the upgrades impact on various subsites. This pull request if far too big for one ticket :)

drupol:
Hi,

Unfortunately I don't have access to that JIRA, this is why I did this pull request.

About tests, it looks like tests are green, are there some other tests that needs to be done ?

Fefaine:
Depending on the module, all test that are necessary and are not yet covered by automated tests

PR #646, opened 2016-06-08
NEXTEUROPA-11356: Adding patch for setting up default timeout for all…

netlooker:
Adding patch. After build no additional tasks are needed to make it work.
This is temporary solution to process with 500 errors investigation.


jfhovinne:
:+1:

jfhovinne:
Patch looks good and applies correctly. Not tested.

voidtek:
:+1:

voidtek:
Not tested in VPS.

degliwe:
:+1:

PR #652, opened 2016-06-09
NEXTEUROPA-11451: fix if the entity isn't a node

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-11451


degliwe:
:+1:

tatarbj:
even we know *normally* nodes have the language property, it should be checked instead of just checking the #node element of $form array.

jfhovinne:
:+1:

PR #659, opened 2016-06-15
Feature/nexteuropa 11506

Fefaine:
Enable entityreference_prepopulate


Fefaine:
Not sure that is the right way to go. Waiting the results of the audit from @cyberwolf

PR #664, opened 2016-06-16
Bugfix/nexteuropa 11426

Fefaine:


tatarbj:
@Fefaine are you sure from ldap (or make it generic: everywhere) the original data is stored in ISO-8859-1. Maybe to use https://api.drupal.org/api/drupal/includes%21unicode.inc/function/drupal_convert_to_utf8/7.x is a bit safer for long term.

netlooker:
:+1:

Fefaine:
:+1:

netlooker:
@tatarbj we decided to solve login issue in first place. After that 
we are going to address character coding issue in follow up ticket NEXTEUROPA-11605.

PR #670, opened 2016-06-17
Hotfix/nexteuropa 10281

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-10281

### DIFF
--- a/profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc
+++ b/profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc
@@ -13,8 +13,6 @@ function nexteuropa_multilingual_language_negotiation_provider_callback($languag
     return $language_url;
   }
 
-  module_load_include('inc', 'includes', 'path');
-
   // $_GET['q'] might not be available at this time, because
   // path initialization runs after the language bootstrap phase.
   $check_prefix = nexteuropa_multilingual_language_negotiation_split_suffix(isset($_GET['q']) ? $_GET['q'] : NULL, $languages);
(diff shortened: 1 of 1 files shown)


Fefaine:
Just created a PR to merge 2.1 back in the branch in order to solve behat warning issue

Fefaine:
:+1:

jfhovinne:
:-1: I would filter out that $_GET['q'].

jfhovinne:
:+1:
Did not test but code looks OK.

Fefaine:
:+1:

dxvargas:
:+1:

dxvargas:
:+1:

tatarbj:
will it work always as it is expected? Maybe instead of that new function platform should use https://api.drupal.org/api/drupal/includes%21path.inc/function/drupal_get_path_alias/7.x

dxvargas:
@tatarbj we will still be working on this language detection behaviour. I will add your input to our conversation so that this option can be evaluated.

voidtek:
@tatarbj : Unfortunately, we can't use these Drupal functions in this code level, some modules aren't yet loaded. (for example: **drupal_lookup_path** creates an error when is an content).
Thanks for your interest.

tatarbj:
@voidtek what i've mentioned above is a core functionality, comes from includes/path.inc, so it is for sure already loaded. The only problem could be with this solution what is implemented here when there is more than one source for the same path (also vice versa), the code is not ready for handle it (sometimes those are legacy paths what was not correctly removed before). Actually drupal_lookup_path also comes from drupal core (declared in includes/path.inc as well), so if it throws error, the issue comes from somewhere else what should be solved because it brakes (God saves! 'hacks') the drupal core!
Thanks @dxvargas for the reaction and hopefully the proper way will be used here :)

Fefaine:
@tatarbj  as you can see this is an hotfix, quite urgent.
I suggest we create a refactor ticket. That's the right time as we now have a good test coverage for the URL managment.

voidtek:
Thanks @tatarbj for your feedback and your expertise.

tatarbj:
sure ladies and gentlemen, based on the urgency obviously it could be there, what i advice to be used is like a small improvement, not blocker at all :) thanks for everyone's openness!

PR #678, opened 2016-06-17
Feature/nexteuropa 11581

jfhovinne:
Flexible Purge and Integration rules.


netlooker:
👍

Fefaine:
:+1:

dxvargas:
:+1:

tatarbj:
sorry for my question but all of these rules are going to be enabled manually? because active 'properties' are false everywhere. i hope it won't depend on the webmasters but will be managed by centralised workflow, like rules ui should be disabled on the subsites.

netlooker:
Flexible purge module is already on the production almost since the beginning.
The goal of the ticket was to deliver additional solution for Varnish cache invalidation for sites which request it. Core functionality of the flexible purge module stays untouched.
For the moment there is possibility to turn on those rules via drush on those particular sites which requested that functionality.
We are still under discussion regarding enabling all of them by default.

degliwe:
:+1:

PR #696, opened 2016-06-24
See https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-11210 -…

alexboss:
… improved messages


Fefaine:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-405

PR #702, opened 2016-06-28
NEXTEUROPA-11351: Behat test for checking invalid tokens

netlooker:
This PR includes change of FeatureContext. I've changed clearing watchdog behaviour from the step level to the scenario level. Now watchdog will be cleared before each scenario. That solution is giving us opportunity to interact with watchdog during given scenario.


netlooker:
👍

voidtek:
:+1:

dxvargas:
👍

Fefaine:
:+1:

PR #722, opened 2016-07-05
NEXTEUROPA-12156: Add bean-entity_render_contextual_links_fix-1954950…

voidtek:
NEXTEUROPA-12156: Add bean-entity_render_contextual_links_fix-1954950-7-D7.patch
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-12156


drupol:
Just for the record, link to the drupal.org issue: https://www.drupal.org/node/1954950

drupol:
This has been reopened because we were unable to display the Contextual Links after applying the patch.
We tried with different themes, the expected result didn't succeeded.

drupol:
:+1:

drupol:
Dear @cyberwolf,

The official patch for bean enable the contextual links through the $title_prefix variable and also add the display of the title in the template (to be consistent).
To avoid unexpected behaviors on our subsites, we've overridden the default template in our theme to only display the contextual links and not the title.

cyberwolf:
Ok thanks for the clarification @dxvargas.
I tested it and it seems to work well.

👍

drupol:
You're welcome @Fefaine.

verbruggenalex:
Hey @drupol , if I'm not mistaken only the $title_suffix contains the actual contextual links. I'm not fully sure what $title_prefix and $title_suffix can output. If users did not make use of titles on the beans before I think it's safer to just output what you need:
```
<div class="<?php print $classes; ?> clearfix"<?php print $attributes; ?>>
  <?php if (isset($title_suffix['contextual_links'])): ?>
  <?php print render($title_suffix['contextual_links']); ?>
  <?php endif; ?>
  <div class="content"<?php print $content_attributes; ?>>
    <?php print render($content); ?>
  </div>
</div>
```

This is also the way that display suite handled it before: https://www.drupal.org/node/1032868

Fefaine:
:+1:

imanoleguskiza:
:+1:

PR #742, opened 2016-07-12
Merge master back to branch

Fefaine:
Merge master back to the branch. This fix won't go in 2.1 but only in 2.2 or maybe even 2.3


Fefaine:
:+1:

PR #744, opened 2016-07-13
NEXTEUROPA-9894: Poetry mock

netlooker:
New feature provided via new submodule which is located in tmgmt_poetry module folder.
All available functionalities are available at /admin/poetry_mock.

First important and required step after installation is to set up Poetry specific variable.
Mock is covering all off currently delivered Behat tests.


voidtek:
:+1:

PR #745, opened 2016-07-13
Added dependency on multisite_drupal_toolbox

verbruggenalex:
This dependency is needed because the starterkit (altered for nems) fails on "install-dev" because of a missing dependency: https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-12350


Fefaine:
:+1:

cyberwolf:
👍

PR #753, opened 2016-07-18
NEXTEUROPA-12320: revert the ticket Bugfix/nexteuropa 11340

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-12320


netlooker:
After investigating more details regarding both functions it seems that in our case it is more secure to use loadHTML which allows to get not well-formed data. Problem with loadXML is that it reports an error instead of throwing an exception when the xml is not well formed.
More details [DOMDocument::loadXML](http://php.net/manual/en/domdocument.loadxml.php)

netlooker:
+1

cyberwolf:
👍

PR #754, opened 2016-07-19
NEXTEUROPA-12447 Using prefix instead of language to verify if suffix has been added a…

kugta:
…lready


Fefaine:
Moved to https://github.com/ec-europa/platform-dev/pull/766

voidtek:
:+1:

PR #757, opened 2016-07-19
Bugbix/nexteuropa 11505 xmlsitemap multilanguage with rel alternate

dxvargas:
no extra command


dxvargas:
PR must be to 2.1-support, so close this.

PR #769, opened 2016-07-21
Added an URL wildcard in order to load communities from URLs

bianchim:
MULTISITE-11890


Fefaine:
:-1:

Fefaine:
Here you need to check the argument is numeric
$group = node_load($path[1]);

bianchim:
@Fefaine The check is made just before : 
if ($path[0] !== 'node' || !is_numeric($path[1])) { return NULL; }

PR #771, opened 2016-07-25
NEXTEUROPA-10404: Fix some errors in the produced Poetry HTML

cyberwolf:


Fefaine:
:+1:

profiles/common/modules/custom/tmgmt_poetry/inc/tmgmt_poetry.webservice.inc
```diff
@@ -338,6 +338,13 @@ function _tmgmt_poetry_replace_job_in_content($content, $job, $job_item) {
       if (!$dom->loadHTML($content)) {
         return FALSE;
       }
+
```
cyberwolf: The actual location of this code does not matter. Simplexml provides another way of accessing the underlying DOM and changes are applied on the same in-memory structure. I agree to avoid possible future confusion it's probably best to move the code to where you placed it.

Fefaine:
:+1:

voidtek:
:+1:

PR #778, opened 2016-07-26
Feature/multisite 11835

champcy:


Fefaine:
@champcy  please fix CS Thanks :)

PR #790, opened 2016-08-01
FIx empty label on error validation message

gervasek:
NEXTEUROPA-12553


gillesdeudon:
I close it because it is obsolete; a new patch has been released since then. It will be inserted through another PR

PR #804, opened 2016-08-05
NEXTEUROPA-12579: Remove tmgmt_poetry_test module

Fefaine:


profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.install
```diff
@@ -158,6 +158,9 @@ function tmgmt_poetry_update_7003() {
 function tmgmt_poetry_update_7004() {
```
gillesdeudon: Deleting the module's code source and then making a simple module_disable do not remove the module reference in the code.
In this specific case where we try to keep it simple and the disabling and the removing fo source files, we have 2 options as this module is never used on site/sub-site, playground and production:

1. Keeping the code like that and add in the hook_update a db_delete to delete the module reference with a where clause "name = 'tmgmt_poetry_test";
2. Removing the hook update because the module is only used locally, no need to treat the case where the module is activated
3.

gillesdeudon: After discussions within the team, it is the 2nd options that has been chosen.

cyberwolf:
👍

voidtek:
:+1:

PR #807, opened 2016-08-08
Multisite 8214 map

batigolix:


Fefaine:
CONNECT will create a new PR. Kevin to confirm if this is to be added in core.
Closeing this one, as it's out of date and generating noise in Jenkins

PR #814, opened 2016-08-12
NEXTEUROPA-12357: Add favicon.ico for older browsers

cyberwolf:
Some older browsers still unconditionally look for a favicon.ico. This causes Drupal to be bootstrapped to serve a 404. We want to avoid this, because it unnecessarily consumes server resources.


netlooker:
👍

dxvargas:
👍

PR #819, opened 2016-08-16
Hotfix/multisite 11586

vever001:


profiles/common/modules/custom/media_avportal/includes/MediaAvportalStreamWrapper.inc
```diff
@@ -63,15 +62,13 @@ class MediaAvportalStreamWrapper extends MediaReadOnlyStreamWrapper {
       else {
         $url = 'http://ec.europa.eu/avservices/cfc/video.cfc?method=getVideoDetails&ref=' . $video_id . '&format=json';
         $response = drupal_http_request($url);
-        if ($response->data == 200) {
+
+        if (!isset($response->error)) {
           $video_info = drupal_json_decode($response->data);
```
verbruggenalex: @vever001 , you are right about bug. I just can not understand how more people weren't complaining about this.

Maybe it's also a good idea to check if required attributes (TITLE and THUMB) are present in the data. I do not see any checks in the streamwrapper functions: https://github.com/ec-europa/platform-dev/blob/1555a482e9499efec14c23207bf356e9a5bb6a90/profiles/common/modules/custom/media_avportal/includes/MediaAvportalStreamWrapper.inc#L84 

So I think this needs a little bit more work here. Maybe define the video info for error first, and merge it with the response data?

Just a thought.

vever001: Hello @verbruggenalex,

The title and thumb are not actually required and might be empty in a lot of cases.
See e.g.: http://ec.europa.eu/avservices/cfc/video.cfc?method=getVideoDetails&ref=I053681&format=json
So I would keep it like that. This is the easiest (most simple) fix IMO.

But I did my own review of media_avportal here in case you want to go deeper: https://github.com/vever001/platform-dev/pull/1
That might be out of scope.

Regards,
Hervé

Fefaine:
Replaced with https://github.com/ec-europa/platform-dev/pull/1133

PR #824, opened 2016-08-17
NEXTEUROPA-12741: Remove feature filtering the language options in no…

gillesdeudon:
a simple "drush cc all" is required


Fefaine:
:+1:

PR #835, opened 2016-08-22
Bugfix/nexteuropa 12244 21

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-12244


voidtek:
Continuousphp is green.

PR #839, opened 2016-08-22
NEXTEUROPA-7528 - Restore Left Sidebar behavior

fab01:


yhuard:
Thanks @fab01 

I re-applied your change in #737

PR #854, opened 2016-08-25
Hotfix/NEXTEUROPA-12761

netlooker:
Hotfix for automatic counter assignation.


profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry_mock/src/Mock/PoetryMock.php
```diff
@@ -458,25 +467,45 @@ private static function getDemandeIdFromJobReference($job_reference) {
...
+  /**
+   * Returning properly formatted reference number string.
+   *
+   * @param array $demande_id
+   *   An array with reference elements.
+   *
+   * @return string
+   *   The string with properly formatted reference number.
+   */
+  public static function prepareReferenceNumber($demande_id) {
```
Fefaine: The mock should use the functions provided by the module, not recreate the functions
Here you should use generateRequestId

tests/features/tmgmt/tmgmt_poetry.feature
```diff
@@ -17,6 +17,26 @@ Feature: TMGMT Poetry features
...
+    And I fill in "Body" with "<body>"
+    And I press "Save"
+    And I select "Published" from "state"
+    And I press "Apply"
+    Then I click "Translate" in the "primary_tabs" region
+    And I select the radio button "" with the id "edit-languages-fr"
+    And I press "Request translation"
+    And I press "Submit to translator"
+    And I store the job reference of the translation request page
+    Then the poetry translation service received the translation request
```
Fefaine: I would first add the step where the counter is received, as IRL you can't receive a translation if you have not first received the counter.
You should check also that the counter needs to be received only the first time with part 0, the second time, the part will increase +1.

netlooker: Yes this is true that we can improve that. But for a quick hotfix I think that this is enough and adding full coverage now will postpone delivering of a hotfix which normally don't need to be covered by tests. Let's create follow-up tickets.

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry_mock/src/Mock/PoetryMock.php
netlooker: Mock is mimicking Poetry Web service so it should have own functions as it is a server. Embedding client functions inside mocked server is binding us a bit. Here we have more flexibility.

tests/features/tmgmt/tmgmt_poetry.feature
cyberwolf: Delphine, afaik the counter (or better, sequence, e.g. "NEXT_EUROPA_COUNTER") is send **inside** the translation request. So you need the translation request first :)

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
```diff
@@ -195,4 +195,26 @@ class TMGMTPoetryMockSubContext extends RawDrupalContext implements DrupalSubCon
...
+   * @Then the poetry translation service received initial request with automatic counter
+   */
+  public function assertPoetryTranslationServiceReceivedTheInitialRequest() {
+    $message = '';
+    if (is_object($this->translationRequest['file'])) {
+      $file_object = $this->translationRequest['file'];
+      $message = file_get_contents($file_object->uri);
+    }
+
+    if (strpos($message, PoetryMock::COUNTER_STRING) === FALSE) {
```
cyberwolf: This check is a bit 'wide'. Would it be feasible to look for the exact element?
The fact is, this test probably succeeds without problems without the fix, because we were sending the sequence value inside the "numero" element while it should be in the "sequence" element. At first sight this code will not catch this mistake.

tests/features/tmgmt/tmgmt_poetry.feature
Fefaine: My understanding is that the request happens in that sequence: (generateRequestId)
- IF You find a numero in the db poetry_map: use the numero only if part is < than 99
- if not (part is 99 or you never had a numero) : request a numero
{code}
// If we have found nothing, request a new numero.
    $id_data = array(
      'codeDemandeur' => $settings['code'],
      'annee' => date("Y"),
      'numero' => $settings['counter'],
      'version' => 00,
      'partie' => 00,
      'produit' => 'TRA',
   );
{code}
Then with this request, you now have a numero and can switch to step 1

netlooker: And you are completely right. The only thing is that this functionality is covered in poetry client implementation and to cover it with test we should create a ticket. There is more things that we can check on poetry client part thus I would like to propose follow up ticket to cover more of them. At that time tmgmt_poetry client was not covered at all thus it is a technical debt that we need to cover.

cyberwolf:
👍

cyberwolf:
👍

cyberwolf:
👍

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.install
```diff
@@ -48,3 +50,26 @@ function tmgmt_poetry_mock_uninstall() {
...
+    foreach ($job_items as $item) {
+      $item->delete();
+    }
+    $job->delete();
+  }
+
+  // Deleting files which were created by tmgmt_poetry_mock.
+  PoetryMock::removeAllRequestTranslationFiles();
+  file_unmanaged_delete_recursive(TMGMT_POETRY_MOCK_REQUESTS_PATH);
+
```
dxvargas: I don't really agree that, by disabling the module, the poetry tables should be truncated. I see more problems than advantages with this. One problem is someone some day tries this module in production but as this module is for testing proposes and it is hidden, may be it is an acceptable risk (although it would have disastrous impacts).
But even for development proposes... do I really want to truncate the poetry tables if I disable this module? And couldn't it be confusing that this occurs silently, without me asking or being informed? Imagine I got a dump from production and at some point I disable the mock...

Fefaine:
:-1:

Fefaine:
Functionality is still broken. See my steps in the comments.

Fefaine:
:+1:

voidtek:
:+1:

Fefaine:
:+1:

PR #855, opened 2016-08-25
Feature/nexteuropa 11174

jfhovinne:


tests/features/text_formats.feature
```diff
@@ -0,0 +1,29 @@
...
+  @api
+  Scenario: Text formats should be available
+    When I go to "admin/config/content/formats"
+    Then I should see "Full HTML"
+    And I should see "Filtered HTML"
+    And I should see "Basic HTML"
+    And I should see "Plain text"
+  
+  @api
+  Scenario: Text formats should be correctly configured
```
cyberwolf: I'd rather like to see a scenario from the standpoint of the editor and the behaviour he expects. The current scenario just does a one-on-one comparison of what's supplied in the hook_filter_default_formats() implementation and the corresponding setting as viewed in the admin UI. We do not test the actual behavior, which is that a <abbr> tag including its title - will be rendered when an editor inserts it in a node body field with one of those filter formats and then views the node.

Fefaine: Editors have not right to change the filter in the standard install. I think it's important we clarify who can do what....

jfhovinne: So the scenario to test must be clarified.

cyberwolf:
👍 
regardless of the comments regarding the test, it's good enough and the chance of breaking this is low IMHO

ademarco:
👍

tests/features/content_editing_editor.feature
```diff
@@ -0,0 +1,40 @@
...
+      | <p><cite>Lorem Ipsum</cite> dolor <abbr title=\"Software Integration Test\">SIT</abbr> amet.</p>    | <p><cite>Lorem Ipsum</cite> dolor <abbr title=\"Software Integration Test\">SIT</abbr> amet.</p>    |
+      | <blockquote>Lorem ipsum dolor sit amet.</blockquote>                                                | <blockquote><p>Lorem ipsum dolor sit amet.</p></blockquote>                                         |
+      | <p>Lorem ipsum <code>dolor(sit);</code> amet.</p>                                                   | <p>Lorem ipsum <code>dolor(sit);</code> amet.</p>                                                   |
+      | <ul><li>Lorem ipsum</li><li>Dolor sit amet</li></ul>                                                | <ul><li>Lorem ipsum</li>                                                                            |
+      | <ol><li>Lorem ipsum</li><li>Dolor sit amet</li></ol>                                                | <ol><li>Lorem ipsum</li>                                                                            |
```
voidtek: A example with "<dl><dt><dd>" is missing, but it's not mandatory and not specify in the ticket.

jfhovinne: Indeed @voidtek 
As explained on the ticket, apparently the "dl dt dd" elements are not properly rendered after saving, so the test would fail.

voidtek:
:+1:

netlooker:
👍

PR #857, opened 2016-08-26
Hotfix/nexteuropa 8719

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-8719

### DIFF
--- a/tests/features/taxonomy.feature
+++ b/tests/features/taxonomy.feature
@@ -6,11 +6,11 @@ Feature: Taxonomy
   @api
   Scenario: Administrator user creates a new vocabulary and adds a new group and terms.
     Given I am logged in as a user with the 'administrator' role
-    When I create a new vocabulary "vocabulary_test"
+    And I create a new vocabulary "vocabulary_test"
     And I create a new group "tab" named "group_test" in the vocabulary "vocabulary_test"
     And I create a new field "text_long" named "field_test" grouped in "group_test" in the vocabulary "vocabulary_test"
     And I create a new term "term_test" in the vocabulary "vocabulary_test"
-    And I go to "admin/structure/taxonomy/vocabulary_test"
+    When I go to "admin/structure/taxonomy/vocabulary_test"
     And I click "Add term"
     And I fill in "Name" with "Term Test 2"
     And I press the "Save" button
--- a/tests/features/taxonomy.feature
+++ b/tests/features/taxonomy.feature
@@ -6,10 +6,10 @@ Feature: Taxonomy
   @api
   Scenario: Administrator user creates a new vocabulary and adds a new group and terms.
     Given I am logged in as a user with the 'administrator' role
-    And I create a new vocabulary "Vocabulary Test"
-    And I create a new group "Vertical tab" named "Group Test" in the vocabulary "Vocabulary Test"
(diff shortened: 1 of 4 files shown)


tests/src/Context/TaxonomyContext.php
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Revert to previous settings after scenario execution.
+   *
+   * @AfterScenario
+   */
+  public function removeVocabularies() {
+    // Remove the vocabularies.
+    foreach ($this->vocabularies as $vocabulary_name) {
+      $vid = db_query("SELECT vid FROM {taxonomy_vocabulary} WHERE machine_name = '$vocabulary_name'")->fetchField();
```
gillesdeudon: I would user taxonomy functions in order to retrieve the vid like:
$vid = taxonomy_vocabulary_machine_name_load('your_vocabulary_name')->vid;
If it is too heavy, I suggest at least to use db_select instead of db_query.

tests/src/Context/TaxonomyContext.php
```diff
@@ -0,0 +1,189 @@
...
+   * @param string $term_name
+   *    Name of the term.
+   * @param string $vocabulary_name
+   *    Name of the vocabulary.
+   *
+   * @Then I create a new term :term_name in the vocabulary :vocabulary_name
+   */
+  public function iCreateNewTermInTheVocabulary($term_name, $vocabulary_name) {
+    // Get the vocabulary ID.
+    $vid = db_query("SELECT vid FROM {taxonomy_vocabulary} WHERE machine_name = '$vocabulary_name'")->fetchField();
```
gillesdeudon: Second time where this query is done, a helper function combined with Taxonomy functions can simplify the code.

tests/features/taxonomy.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Taxonomy
+  In order to manage the taxonomy on the website
+  As an administrator
+  I want to be able to create, edit and delete taxonomy
+
+  @api
+  Scenario: Administrator user creates a new vocabulary and adds a new group and terms.
+    Given I am logged in as a user with the 'administrator' role
+    And I create a new vocabulary "vocabulary_test"
+    And I create a new group "tab" named "group_test" in the vocabulary "vocabulary_test"
```
gillesdeudon: Instead of machine names, we should use label as much as possible. 

For instance instead of "tab", I rather use "Vertical tab" instead of "tab" that is meaningful for business understanding. In the function treating the step, the field group label could be converted into its machine name by a helper function.

Other example, for field and field group, implement the same logic as what exist in the "Manage fields" interface where by default, the machine name is automatically defined from the field label .

tests/src/Context/TaxonomyContext.php
```diff
@@ -0,0 +1,275 @@
...
+  /**
+   * Create a machine name.
+   *
+   * @param string $name
+   *    Name.
+   *
+   * @return string
+   *    Machine name.
+   */
+  private function getMachineName($name) {
```
jfhovinne: Would it be possible to use an existing function for this?

jfhovinne: OK, we agreed to improve this later on if needed.

jfhovinne:
:+1:

Fefaine:
:+1:

PR #868, opened 2016-09-01
fix(contributing): Fix the fatal error when creating a newsroom term.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-8719


jfhovinne:
:+1:

voidtek:
:+1:

PR #871, opened 2016-09-01
fix(multilingualism): Fix translations are not submitted when the counter doesn't exist

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-12759

On the branch Master.


Fefaine:
:+1:

netlooker:
👍

PR #872, opened 2016-09-02
feature(interoperability): add web frontend cache purge rules

cyberwolf:


voidtek:
:+1:

netlooker:
👍

tatarbj:
@cyberwolf could you contact with Gregory, last time when we saw curl implementation by a subsite he said they should use drupal_http_request because it is handled by ec proxy things, he knows it better but maybe if it will be deployed it won't fully work because of curl.

cyberwolf:
Don't worry, it is based on existing code. I tested it on playground and it works as expected.

PR #875, opened 2016-09-05
fix(multilingual): do not use und in alias for i18n contents

dxvargas:
no additional commands needed

### DIFF
--- a/profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
+++ b/profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
@@ -419,7 +419,7 @@ function nexteuropa_multilingual_path_update($path) {
 /**
  * Force the path to have a language neutral alias.
  *
- * If entity does not have a translation mechanism other than entity translation
+ * If entity is a node using entity translation,
  * force the path to have a language neutral alias even if it is custom.
  *
  * @param array $path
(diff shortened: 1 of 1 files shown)


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -427,6 +430,16 @@ function nexteuropa_multilingual_path_update($path) {
  *     - language: The language of the alias.
  */
 function _nexteuropa_multilingual_path_und_sync($path) {
+
+  // Check if entity has a translation mechanism other than entity translation.
+  $node = menu_get_object("node", 1, $path['source']);
```
jfhovinne: 1) OK, we are dealing with nodes only. Should perhaps be mentioned in the comments.
2) Are we sure $path['source'] is always set?

profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -495,11 +508,18 @@ function nexteuropa_multilingual_path_delete($path) {
...
+ * If entity does not have a translation mechanism other than entity translation
+ * force all aliases to be saved as language neutral.
+ * All content translations will have a unique alias.
  */
 function nexteuropa_multilingual_pathauto_alias_alter(&$alias, array &$context) {
-  // Force all aliases to be saved as language neutral.
-  // All content translations will have a unique alias.
-  $context['language'] = LANGUAGE_NONE;
+
+  $translation_mechanism = variable_get('language_content_type_' . $context['type'], ENTITY_TRANSLATION_ENABLED);
```
jfhovinne: Are we sure $context['type'] is always set?

jfhovinne:
:+1:

profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
gillesdeudon: Concerning the $path['source'], on working to another ticket, I observed that it was not always set. I do not remember precisely in which conditions. I hesitate between content types that do not use "content translation" nor "Entity translation" or those that do not use revision.

gillesdeudon:
:+1:

jfhovinne:
:+1:

gillesdeudon:
:+1:

PR #879, opened 2016-09-07
NEXTEUROPA-13093: Add patch to workbench moderation to prevent messing with url aliasing

bianchim:


Fefaine:
Patch is now included in workbench_moderation 3.0 we use

PR #890, opened 2016-09-29
NEPT-64: ec-resp / block.tpl.php - missing tag $attributes.

drupol:
This PR fixes two things:

1. The missing tag `$attributes`,
2. The missing comments in many templates in Drupal. See [#569362](https://www.drupal.org/node/569362). I've uploaded a patch upstream, and included in makefile.

You need to flush cache to get this working properly.

Internal issue link: https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-64


cyberwolf:
👍

gillesdeudon:
:+1:

PR #894, opened 2016-09-29
fix(contributing): Fix the internal link token not visible.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-97

### DIFF
--- a/profiles/common/modules/features/cce_basic_config/cce_basic_config.install
+++ b/profiles/common/modules/features/cce_basic_config/cce_basic_config.install
@@ -2219,3 +2219,13 @@ function cce_basic_config_update_7213() {
   // Enable token_filter on full_html text filter.
   multisite_config_service('filter')->setTextFilterWeight('full_html', 'filter_tokens', 20);
 }
+
+/**
+ * NEPT-97.
+ *
+ * Change the filter_tokens weight to be on the bottom of the processing order.
+ */
+function cce_basic_config_update_7214() {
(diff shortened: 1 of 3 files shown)


profiles/common/modules/features/cce_basic_config/cce_basic_config.install
```diff
@@ -2219,3 +2219,13 @@ function cce_basic_config_update_7213() {
...
 }
+
+/**
+ * NEPT-97.
+ *
+ * Change the filter_tokens weight to be on the bottom of the processing order.
+ */
+function cce_basic_config_update_7214() {
+  // Enable token_filter on full_html text filter.
+  multisite_config_service('filter')->setTextFilterWeight('full_html_track', 'filter_tokens', 20);
```
tatarbj: previously 7213 was not fully executed, on the subsites there are a lot of sites (over 70%) without that setting. this one also should be checked after playground deployment with the following drush command:
$ drush @sites sqlq 'select status from filter where format ="full_html_track" and name = "filter_tokens"'
if all sites return with 1 you can be sure the update was fully done.

voidtek: Thank you for your comment, but this PR is in wip, and I prefer your comment during the QA phase because it can create a confusing, now.
Thanks for your understanding.

gillesdeudon:
:+1:

Fefaine:
:+1:

PR #896, opened 2016-09-29
fix(contributing): Fix the empty popup 'Insert internal content'.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-105

### DIFF
--- a/profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
+++ b/profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
@@ -199,6 +199,9 @@ function nexteuropa_token_ckeditor_token_filter_filter_tokens($text, $filter, $f
 
 /**
  * Implements hook_custom_theme().
+ *
+ * NEPT-105:The popup 'Insert internal content' (on a Ckeditor block)
+ * when the theme of the modal and the back page isn't the same.
  */
 function nexteuropa_token_ckeditor_custom_theme() {
   if (strpos(current_path(), 'nexteuropa/token-ckeditor') === 0) {
--- a/profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
+++ b/profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
@@ -201,10 +201,10 @@ function nexteuropa_token_ckeditor_token_filter_filter_tokens($text, $filter, $f
  * Implements hook_custom_theme().
  */
 function nexteuropa_token_ckeditor_custom_theme() {
-  $theme = variable_get('node_admin_theme', '1') ?
-    variable_get('admin_theme_default', 'seven') :
-    variable_get('theme_default', 'ec_resp');
   if (strpos(current_path(), 'nexteuropa/token-ckeditor') === 0) {
+    $theme = variable_get('node_admin_theme', '0') ?
+      variable_get('admin_theme_default', 'seven') :
+      variable_get('theme_default', 'ec_resp');
     return $theme;
   }
 }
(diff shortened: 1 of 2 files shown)


tatarbj:
:-1:

profiles/common/modules/features/nexteuropa_webtools/nexteuropa_webtools.module
```diff
@@ -268,3 +268,15 @@ function nexteuropa_webtools_ajax_access_callback() {
   $view = views_get_view($view_name);
   return $view && $view->access($display_id);
 }
+
+/**
+ * Implements hook_custom_theme().
+ */
+function nexteuropa_token_ckeditor_custom_theme() {
```
tatarbj: redeclared function name. please do not copy-paste code.

tatarbj:
btw why that codacy/pr says 'Good work!' ?! there is a syntax error...

profiles/common/modules/features/nexteuropa_webtools/nexteuropa_webtools.module
voidtek: it's in work in progress.

tatarbj:
seems ok for me, but still not enough for fixing the issues caused by 2.2.133 deployment on production.

voidtek:
We have an other PR.
https://github.com/ec-europa/platform-dev/pull/894

profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
```diff
@@ -196,3 +196,15 @@ function nexteuropa_token_ckeditor_token_filter_filter_tokens($text, $filter, $f
...
 }
+
+/**
+ * Implements hook_custom_theme().
+ */
+function nexteuropa_token_ckeditor_custom_theme() {
+  if (strpos(current_path(), 'nexteuropa/token-ckeditor') === 0) {
+    $theme = variable_get('node_admin_theme', '0') ?
+      variable_get('admin_theme_default', 'seven') :
+      variable_get('theme_default', 'ec_resp');
```
gillesdeudon: Before the PR 570, the code here was 
`if (strpos(current_path(), 'nexteuropa/token-ckeditor') === 0) {
    return 'seven';
  }`

The use of this value `variable_get('admin_theme_default', 'seven')` does not bother me so much because it is commonly used in all sites but `variable_get('theme_default', 'ec_resp')` bothers me because we cannot be sure that the default theme is compatible with the feature or Am I wrong?

profiles/common/modules/features/nexteuropa_webtools/nexteuropa_webtools.module
```diff
@@ -268,3 +268,15 @@ function nexteuropa_webtools_ajax_access_callback() {
...
 }
+
+/**
+ * Implements hook_custom_theme().
+ */
+function nexteuropa_webtools_ckeditor_custom_theme() {
+  if (strpos(current_path(), 'nexteuropa/webtools-ckeditor') === 0) {
+    $theme = variable_get('node_admin_theme', '0') ?
+      variable_get('admin_theme_default', 'seven') :
+      variable_get('theme_default', 'ec_resp');
```
gillesdeudon: Before the PR 570, this code is inspired by "nexteuropa_token_ckeditor_custom_theme" where  we had: 
`if (strpos(current_path(), 'nexteuropa/token-ckeditor') === 0) {
    return 'seven';
  }`

The use of this value `variable_get('admin_theme_default', 'seven')` does not bother me so much because it is commonly used in all sites but `variable_get('theme_default', 'ec_resp')` bothers me because we cannot be sure that the default theme is compatible with the feature or Am I wrong?

netlooker:
👎  More information in the ticket.

netlooker:
👍

gillesdeudon:
I agree concerning the behat test but as it is a hotfix and we havea problem with PhatomJS in cPHP, we should do it through a follow up in order to not block the ticket

gillesdeudon:
:+1:

profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module
tatarbj: @gillesdeudon actually if i'm not wrong ec_resp is currently the base theme if you install a new NE instance, i see it here: https://github.com/ec-europa/platform-dev/blob/d7e625e773df942543a8d970c9151a27bbed584d/profiles/common/modules/features/multisite_settings/multisite_settings_core/multisite_settings_core.strongarm.inc#L140 and because multisite_settings_core is a dependency of multisite_drupal_standard profile (https://github.com/ec-europa/platform-dev/blob/d7e625e773df942543a8d970c9151a27bbed584d/profiles/multisite_drupal_standard/multisite_drupal_standard.info#L118) you can be sure this theme is currently the base theme on every subsites. The usage of variable_get is required like this because if something else is the theme_default, it has to be used, otherwise the 'common' one: ec_resp -- is it clear like this? :)

PR #897, opened 2016-09-30
NEPT-310: This prevent the build to fail on some newer virtual machines.

drupol:
Actually the drush make workflow automatically apply patches to projects.
In some cases, it wont work properly due to some version of the git and patch softwares.
This patch fix the problem by using the right standard patch that works everywhere.

Please read more information here: https://www.drupal.org/node/1903850

### DIFF
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -526,7 +526,9 @@ projects[piwik][subdir] = "contrib"
 projects[piwik][version] = "2.9"
 
 projects[plupload][subdir] = "contrib"
-projects[plupload][version] = "1.7"
+projects[plupload][download][branch] = 7.x-1.x
+projects[plupload][download][revision] = bba974c6f3224346a1acae4181a700b55129e6e1
+projects[plupload][download][type] = git
 
 projects[print][subdir] = "contrib"
 projects[print][version] = "2.0"
@@ -897,6 +899,15 @@ libraries[leaflet][destination] = "libraries"
 libraries[leaflet][download][type] = "file"
 libraries[leaflet][download][url] = "http://cdn.leafletjs.com/downloads/leaflet-0.7.5.zip"
 libraries[leaflet][directory_name] = "leaflet"
+
+; Plupload
+libraries[plupload][destination] = "libraries"
+libraries[plupload][download][type] = "file"
+libraries[plupload][download][request_type]= "get"
+libraries[plupload][download][file_type] = "zip"
+libraries[plupload][download][url] = "https://github.com/moxiecode/plupload/archive/v1.5.8.zip"
+libraries[plupload][directory_name] = "plupload"
+libraries[plupload][patch][1903850] = "https://www.drupal.org/files/issues/plupload-1_5_8-rm_examples-1903850-21.patch"
 
 ; ===========================
 ; Libraries for Ec_resp Theme
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -525,6 +525,13 @@ projects[pathauto_persist][version] = "1.4"
 projects[piwik][subdir] = "contrib"
 projects[piwik][version] = "2.9"
 
+; Instead of using a stable version of the plupload module, we stick here to a
+; more recent git revision in order to solve unexpected failures with a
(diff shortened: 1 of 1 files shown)


resources/multisite_drupal_standard.make
```diff
@@ -526,7 +526,9 @@ projects[piwik][subdir] = "contrib"
 projects[piwik][version] = "2.9"
 
 projects[plupload][subdir] = "contrib"
-projects[plupload][version] = "1.7"
+projects[plupload][download][branch] = 7.x-1.x
+projects[plupload][download][revision] = bba974c6f3224346a1acae4181a700b55129e6e1
```
cyberwolf: @drupol can you document this section of the make file, describing why we use the particular revision? It will be easier then for our future selfs and for others to know why this was changed.

resources/multisite_drupal_standard.make
```diff
@@ -890,6 +892,15 @@ libraries[leaflet][download][type] = "file"
...
 libraries[leaflet][directory_name] = "leaflet"
 
+; Plupload
+libraries[plupload][destination] = "libraries"
+libraries[plupload][download][type] = "file"
+libraries[plupload][download][request_type]= "get"
+libraries[plupload][download][file_type] = "zip"
+libraries[plupload][download][url] = "https://github.com/moxiecode/plupload/archive/v1.5.8.zip"
+libraries[plupload][directory_name] = "plupload"
+libraries[plupload][patch][1903850] = "https://www.drupal.org/files/issues/plupload-1_5_8-rm_examples-1903850-29.patch"
```
cyberwolf: Can you add a small description and the URL to the issue?

drupol:
Hi,

I'm off until Wednesday, I will try to take care of it as home asap.

Thanks.

cyberwolf:
👍

gillesdeudon:
I cannot give my thumb-up for this PR because pointing to a specific commit of the Dev versions introduce some uncertainties for the future. I will create a ticket specific to the problem met by Pol and attached this PR to it.

gillesdeudon:
:-1:

drupol:
When you say "uncertainties", what do you mean ? The version is fixed and will never change.
We've reviewed the changes from the stable version to this specific version with @cyberwolf already.
There were almost no changes that worth being suspicious at all.

gillesdeudon:
The "almost" of your comment is already an uncertainty. With the 1.7 of Plupload, no ticket has been raised from sites. I have no guarantee that still will be the case with this Dev revision commit.

So, I prefer that we treat this problem separately from the NEPT-64 as it has no relationship with the issue and we investigate why the issue happens only on one environment and not on all first before using a specific Dev commit and risking future problems. 
That's why I create a ticket for this topic and I put NEPT-64 in acceptance or even in Done.

resources/multisite_drupal_standard.make
drupol: Done by @cyberwolf .

resources/multisite_drupal_standard.make
drupol: Done by @cyberwolf .

drupol:
We reviewed the changes and they won't impact the platform.

Can we push this ticket forward ?

CC @jfhovinne @Fefaine @cyberwolf

Fefaine:
@drupol what do you mean "we reviewed' ? You mean QA?
Anyway, i think this PR should be closed as the linked ticket is closed.
If you have another ticket, then create a new PR for this ticket (if it's in the sprint). If not, create a ticket and wait for it to be added to the sprint please.

ademarco:
👍

ademarco:
Setting a specific commit hash is removing uncertainty as it is the same as using a tag (tags are just labels on commit hashes at the end).

Fefaine:
imo this should not go to master, but instead to 2.3

Fefaine:
:-1:

drupol:
This is a blocker and this should go to master as the changes in the code are not impacting anything.

FYI, [here a link that](http://pastebin.com/wWLLXUfg) shows the diff between the version currently used (1.7) and the version I would like to use.

Please, could you think about it twice?

drupol:
Closing this in favor of: #1072

PR #898, opened 2016-10-03
fix(multilingual) Add multisite_translations as a dependency for tmgm…

Fefaine:
…t_poetry


Fefaine:
See related ticket, request was made in error

PR #901, opened 2016-10-04
fix(multilingualism): Reject all jobs when the translation is rejected by a poetry configuration error.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-55


netlooker:
Together with [~costvit] we decided to create a new PR to merge outcome from other tickets related to NEPT-102.
The final PR is available here - https://github.com/ec-europa/platform-dev/pull/912

PR #907, opened 2016-10-06
NEPT-60: TMGMT: Fix insufficient access check on Views.

drupol:
Some Views do not provide sufficient access check, which is considered as security risk , as it may allow to expose the view information to the unwanted user.

This PR fix the problem by adding an upstream patch to the makefile.

Link to the drupal.org issue: https://www.drupal.org/node/2812863
JIRA Link: https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-60


dxvargas:
👍

verbruggenalex:
:+1:

PR #908, opened 2016-10-07
feat(interoperability): trigger cache purge rules as well on non-moderated content

cyberwolf:
Implements NEPT-87.

### DIFF
--- a/tests/src/Context/DrupalContext.php
+++ b/tests/src/Context/DrupalContext.php
@@ -65,7 +65,14 @@ class DrupalContext extends DrupalExtensionDrupalContext {
   public function rememberCurrentLastNode() {
     $query = db_select('node');
     $query->addExpression('MAX(nid)');
-    $this->maxNodeId = $query->execute()->fetchField();
+    $max_node_id = $query->execute()->fetchField();
+
+    if (NULL === $max_node_id) {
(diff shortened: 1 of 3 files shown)


tests/src/Context/DrupalContext.php
```diff
@@ -50,4 +57,35 @@ public function visitContentPage($type, $title) {
...
+    $this->maxNodeId = $query->execute()->fetchField();
+  }
+
+  /**
+   * Removes any nodes created after the last node id remembered.
+   *
+   * @AfterScenario @reset-nodes
+   */
+  public function resetNodes() {
+    $max_nid = isset($this->maxNodeId) ? $this->maxNodeId : 0;
```
ademarco: By setting this to 0 if not set we will delete nodes in the system, right? The method name is a bit misleading then, since I would expect it to reset my site at the state before the scenario execution.

Anyway I would not delete all nodes but only those created in the scenario, the platform comes with node 1 which is created at installation stage (the default editorial group) and this would delete that node, which is annoying for developers mainly.

ademarco:
👍

tatarbj:
:+1:

gboddin:
Can I know which caches will be purged in which conditions ? That sounds quite risky to me

PR #913, opened 2016-10-11
feat(interoperability): use limited wildcard patterns in purge path

cyberwolf:


netlooker:
We've noticed some random problems with CPHP. The code locally works without issues and tests are green.

netlooker:
👍

tatarbj:
:+1:

PR #915, opened 2016-10-11
NEPT-183: feat(maintenance) Upgrade drupal core to 7.51

Fefaine:


drupol:
Here's the updated patch.

[default-settings-php-include-local-settings-drupal-751.patch.txt](https://github.com/ec-europa/platform-dev/files/522744/default-settings-php-include-local-settings-drupal-751.patch.txt)

drupol:
I think we should document those sections in the file.

drupol:
Should we include the settings override on the top of the file ?
What is the purpose of it ?
If it's meant to override, why it's not, by default, on the bottom of the file ?

@cyberwolf ? @jfhovinne ? Your point of view ?

drupol:
There's another issue.

[The patch @pfrenssen created](https://www.drupal.org/node/1399846) is conflicting with a new hook_update_N() in Drupal 7.51: the [system_update_7082()](http://cgit.drupalcode.org/drupal/tree/modules/system/system.install?h=7.51#n3276).

We are investigating on this to check what is the best way to fix this.

drupol:
Right now, we are using a patch without the [hook_update_N() that was in the patch](https://www.drupal.org/node/1399846?page=1#comment-11718181).

A better solution would be to backport this in D7: https://www.drupal.org/node/2801777

gillesdeudon:
@drupol, I will make some test on the "backport" patch you made on  https://www.drupal.org/node/2801777 but if I have a doubt about its stability, I will stay on your first patch and create a specific ticket for https://www.drupal.org/node/2801777 because I think it deserve its own ticket.

drupol:
Sure, let's do that.

drupol:
CPHP failed again.

imanoleguskiza:
I tested the patch you proposed and it works fine. Both the simpletest and behat test were fine in my local environment.

However this new backport patch replaces a whole chunk of functionality: the ability to override the maximum age for files (set to 21600 by default). So before deciding to use this patch I guess we should try to assess the impact of it (is any site even using this feature?).

PR #916, opened 2016-10-12
perf(interoperability): do not enable apachesolr_multisitesearch by default

cyberwolf:
Implements NEPT-173.


netlooker:
👍

verbruggenalex:
Commented on the QA ticket: https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-13587
I don't have any problems with the pull request itself except that the dependency should also be removed from the communities profile:
https://github.com/ec-europa/platform-dev/blob/master/profiles/multisite_drupal_communities/multisite_drupal_communities.info#L9

PR #917, opened 2016-10-12
perf(interoperability): do not enable apachesolr and its friends by default

cyberwolf:
Implements NEPT-174.


netlooker:
👍

profiles/common/modules/features/communities/communities.install
```diff
@@ -18,11 +18,14 @@ function communities_disable() {
...
-    'bundle' => 'community',
-  ))->execute();
+  if (module_exists('apachesolr')) {
+    db_insert('apachesolr_index_bundles')->fields(array(
+      'env_id' => 'solr',
+      'entity_type' => 'node',
+      'bundle' => 'community',
+    ))->execute();
+  }
+
```
tatarbj: you should also remove the dependencies from community profile after watching this change, shouldn't you?

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
```diff
@@ -803,9 +806,24 @@ function _disable_block($module, $delta) {
 /**
  * Public function to add or delete nodes from the index.
  *
- * You need to add the module apachesolr as a dependancy to your feature.
+ * You need to add the module apachesolr as a dependency to your feature.
  */
 function multisite_drupal_toolbox_config_solr_bundle($bundle, $op) {
```
tatarbj: Please complete the header documentation for this function with the formal parameter list's items.

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
```diff
@@ -803,9 +806,24 @@ function _disable_block($module, $delta) {
...
+  if (!module_exists('apachesolr')) {
+    watchdog(
+      'multisite_drupal_toolbox',
+      'Function %function was called with bundle %bundle and op %op, but the required apachesolr module is not enabled. You either need to add the module apachesolr as a dependency to the feature you are calling this function from, or prevent the call from happening if apachesolr is not a strict dependency and is disabled.',
```
tatarbj: make it translated https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7.x

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
cyberwolf: This is out of scope of the issue IMHO.

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
cyberwolf: @tatarbj watchdog() actually performs the translation. See its documentation:

> $message: The message to store in the log. Keep $message translatable by not concatenating dynamic values into it! Variables in the message should be added by using placeholder strings alongside the variables argument to declare the value of the placeholders. See t() for documentation on how $message and $variables interact.

profiles/common/modules/features/communities/communities.install
cyberwolf: @tatarbj the community profile is not maintained by the core team.

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
tatarbj: actually it asks for t() :) but also throws a phpcs error that was discovered by @voidtek so for now i would say it's ok.

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
tatarbj: great, let's skip a 2 minutes issue ;)

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
cyberwolf: @tatarbj no there should not be a t() here at all. t() needs to be used e.g. where the entries of watchdog are printed on the screen by the dblog module, see for example https://api.drupal.org/api/drupal/modules!dblog!dblog.admin.inc/function/theme_dblog_message/7.x. If you use t() where you suggest it now, the log message will always show up in the current language when the message was logged (given there is a translation of course) instead of the current language when the message is displayed.

tatarbj:
👍

PR #918, opened 2016-10-13
style(site template): do not print class attribute of a link when it is empty

cyberwolf:
Implements NEPT-156.


drupol:
:+1:

verbruggenalex:
:+1: All good.

PR #924, opened 2016-10-13
fix(contributing): Apply contrib patch on the module wysiwyg '2414575'.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1


gillesdeudon:
:+1:

tatarbj:
:+1:

PR #927, opened 2016-10-13
Added PhantomJS restart before each scenario & Phing target to stop unused services

ppaulis:
The build timeouts we had in the past, were generated through random freezes of PhantomJS. By adding a BeforeScenario hook in Behat that restarts PhantomJS, the problem should be solved.

Additionally, there is now a Phing target "stop-services" that stops unused services in the container to free additional memory. (After having accepted the PR, this target has to be set as the 1st target in the list of Phing targets in the continuousphp Behat config).

Greetings,
Pascal


cyberwolf:
@ppaulis the changes you made seem to cause other failures, so I think this needs more work.

PR #932, opened 2016-10-14
fix(multilingual): fix error when content has language UND

dxvargas:
The token "**node:entity-translation-languages**" and the variable "**nexteuropa_multilingual_warning_message_languages**" will be removed from code. But maybe they are currently being used in custom code in some subsites.
The removal of the variable should not cause any problem, because it will still exist in the database (**no** hook_update_N with variable_del() is created).
But the removal of the token will make the token not to be replaced. This is not a critical error, but still if possible in the QA process by the maintenance team, some evaluation of the uses of this token in subsites can have place. If considered too risky, we can reconsider the removal of this token.


cyberwolf:
👍

verbruggenalex:
:+1:

PR #933, opened 2016-10-14
feat(multilingualism): 'Requested delivery date' field improvements

netlooker:


cyberwolf:
Looks good.
Please note I've tested locally wether the hardcoded dates won't cause any issues when they are in the past or outside of the year range (they will be in the future). This did not cause the tests to fail, the date field does not validate this currently, and this was not in scope of this ticket.

cyberwolf:
👍

drupol:
:rainbow:

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -205,6 +205,15 @@ function _tmgmt_poetry_form_load_translator_settings(&$form_state, &$form, $job)
     $settings = [];
   }
 
+  // NEPT-62: Overwriting the 'delai' field properties.
+  if (isset($settings['delai'])) {
+    $settings['delai']['#title'] = t('Requested delivery date');
+    $settings['delai']['#date_year_range'] = '-0:+1';
```
tatarbj: originally it's -3;+3 https://github.com/ec-europa/platform-dev/blob/a8b9d9625be4d6109a05cb03430edad89e70d37e/profiles/common/modules/custom/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc#L310 why you change it here? the ticket is about making it required only, isn't it?

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -205,6 +205,15 @@ function _tmgmt_poetry_form_load_translator_settings(&$form_state, &$form, $job)
...
+    $settings['delai']['#default_value'] = '';
```
tatarbj: also the default value is set previously and you overwrite it here to an empty string!

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -205,6 +205,15 @@ function _tmgmt_poetry_form_load_translator_settings(&$form_state, &$form, $job)
...
+    $settings['delai']['#required'] = TRUE;
+    $settings['delai']['#weight'] = -10;
```
tatarbj: the ticket was not about making it in the top with setting up the #weight (it doesn't matter that much, but still the change makes more than it's expected!)

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
Fefaine: You cannot request a translation in the past.

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
Fefaine: Yes, the ticket is to force the user to choose a date in the future, not provide one by default

tatarbj: ok

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
tatarbj: so it didn't change, you kept it, ok for me.

tatarbj:
👍

profiles/common/modules/custom/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -205,6 +205,15 @@ function _tmgmt_poetry_form_load_translator_settings(&$form_state, &$form, $job)
     $settings = [];
   }
 
+  // NEPT-62: Overwriting the 'delai' field properties.
+  if (isset($settings['delai'])) {
```
imanoleguskiza: I don't really understand why we overwrite the field properties when we are actually defining the form ourselves. Shouldn't this change be generic for any request?

Fefaine: Actually we want to override the label. The date can be changed by DGT, and the expected delivery date won't be the requested delivery date if that is changed.
So if it was set, we change the label to differenciate this with the date that will show in the jobs

profiles/common/modules/custom/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc
```diff
@@ -312,9 +312,11 @@ class TMGMTPoetryTranslatorUIController extends TMGMTDefaultTranslatorUIControll
     $form['delai'] = array(
       '#type' => 'date_popup',
       '#title' => t('Requested delivery date'),
-      '#default_value' => (!empty($job->settings['delai']) ? $job->settings['delai'] : date('Y-m-d', strtotime('+1 month'))),
+      '#default_value' => '',
       '#date_format' => 'd/m/Y',
-      '#date_year_range' => '-3:+3',
+      '#date_year_range' => '-0:+1',
+      '#required' => TRUE,
+      '#weight' => -10,
```
voidtek: Was weight change included in the scope?

profiles/common/modules/custom/tmgmt_poetry/src/tmgmt_poetry.behat.inc
```diff
@@ -228,6 +228,16 @@ class TmgmtPoetrySubContext extends RawDrupalContext implements DrupalSubContext
   }
 
   /**
+   * Generate a relative date for job deadline.
+   *
+   * @When I fill in :arg1 with a relative date of :arg2 days
+   */
+  public function iFillInWithArelativeDateOfDays($field, $number) {
+    $relative_date = date('d/m/Y', strtotime($number));
+    $this->getSession()->getPage()->fillField($field, $relative_date);
```
voidtek: Could insert here a check:

// Retrieve the specified field.
    if (!$field = $this->getSession()->getPage()->findField($field)) {
      throw new ExpectationException("Field '$field' not found.", $this->getSession());
    }

profiles/common/modules/custom/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc
netlooker: The date field is the only mandatory field that you need to fill. To improve UX after internal discussion we decided to have this field as a first in the field group.

profiles/common/modules/custom/tmgmt_poetry/src/tmgmt_poetry.behat.inc
netlooker: Good point. Thanks for the suggestion.

dxvargas:
👍

dxvargas:
👍

voidtek:
:+1:

dxvargas:
👍

voidtek:
:+1:

netlooker:
I've noticed that one test failed but on local there is no problem and the committed code does not touch it. I consider that as a Jenkins environment issue I will investigate that during the merge of the NEPT-182.
cc: @Fefaine

PR #934, opened 2016-10-14
Add patch for og_field_access hiding fields in content creation

bianchim:
NEXTEUROPA-13743


gervasek:
Close: merge on release/communities before: https://github.com/ec-europa/platform-dev/pull/1013

PR #937, opened 2016-10-18
Feature/nept 163

dxvargas:


drupol:
Hi @dxvargas ,

When reading the description of the ticket, it says that the endpoint URL should be updated too.
I don't see anything related to that in the diff. Could you let me know if this is normal or not ?

ademarco:
👍

dxvargas:
This PR should be deleted because it will be replaced by https://github.com/ec-europa/platform-dev/pull/955

PR #938, opened 2016-10-18
MULTISITE-13557

davidsanchezcrespillo:
Removal of reference to the placeholder theme on ECAS logout

### DIFF
--- a/profiles/common/modules/custom/ecas/includes/ecas.inc
+++ b/profiles/common/modules/custom/ecas/includes/ecas.inc
@@ -213,7 +213,7 @@ function ecas_login_page() {
 function ecas_logout($full_logout = FALSE) {
   global $user;
 
-  watchdog('user', 'Session closed for %name.', array('%name' => theme('placeholder', $user->name)));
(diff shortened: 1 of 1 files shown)


profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -213,7 +213,7 @@ function ecas_login_page() {
 function ecas_logout($full_logout = FALSE) {
   global $user;
 
-  watchdog('user', 'Session closed for %name.', array('%name' => theme('placeholder', $user->name)));
+  watchdog('user', 'Session closed for %name.', array('%name' => $user->name));
```
drupol: I think it's better to use the API function: `format_username()`.

Another PR is here about it: https://github.com/ec-europa/platform-dev/pull/1200

davidsanchezcrespillo: OK for me! I did not know about `format_username()`, it looks a much better alternative.

Fefaine:
PR replaced with https://github.com/ec-europa/platform-dev/pull/1200

PR #940, opened 2016-10-20
feat(interoperability): clear the web frontend cache of particular content when an action is performed on it

cyberwolf:
Implements NEPT-80.


netlooker:
👍

profiles/common/modules/custom/nexteuropa_varnish/src/PurgeRuleType.php
```diff
@@ -0,0 +1,23 @@
...
+ * Enum of purge rule types.
+ */
+class PurgeRuleType {
+  const NODE = 'node';
+  const PATHS = 'paths';
+
+  /**
+   * Prevents the class from instantiated, it just acts as an enum.
+   */
+  private function __construct() {
```
cyberwolf: It's not a singleton. It's a basic lightweight implementation of an enumeration, containing valid purge rule types:

PurgeRuleType::NODE
PurgeRuleType::PATHS

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -151,8 +153,40 @@ function _nexteuropa_varnish_get_paths_to_purge($content_type) {
       'nexteuropa_varnish_cache_purge_rule',
       array_keys($result['nexteuropa_varnish_cache_purge_rule'])
     );
+    /** @var \Drupal\nexteuropa_varnish\Entity\PurgeRule $purge_rule */
```
cyberwolf: Exactly.

tatarbj: great. still against the drupal coding standards.

profiles/common/modules/custom/nexteuropa_varnish/src/PurgeRuleType.php
cyberwolf: @tatarbj I think you are following your own personal opinion and habits, instead of allowing and learning a more modern style of code. I will not change this as I don't agree with your reasoning. There are other and far better approaches than what you are used to in Drupal 7. Time to get used to it with Drupal 8 around the corner.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
cyberwolf: I could entirely remove the line, which will actually be worse towards developers, espcially when you are using a proper IDE.

What is most important? Following the exact rules like they are written down like a blind follower - our code sniffer even did not fail on it AFAIK - or improve code readability and automatic introspection by IDEs?

tatarbj:
The refactored code makes much more sense than it was before, thanks for the work, for me the PR is acceptable now.

tatarbj:
:+1:

PR #941, opened 2016-10-21
feat(multilingualism): add website identifier to title of translation requests

cyberwolf:


tests/features/tmgmt/tmgmt_poetry.feature
```diff
@@ -417,3 +417,30 @@ Feature: TMGMT Poetry features
...
       And I press "Submit to translator"
       Then I see the date of the last change in the "French" row
+
+  # Deliberately not using a JavaScript enabled browser here, as it will probably
+  # respect the maximum length specified on the input field and automatically
+  # trim any value we fill it with.
+  Scenario: A website identifier longer than 15 characters is not accepted.
+    When I go to "/admin/config"
+    And I click "DGT Connector"
+    And I fill in "itphachyufbabtap" for "Website identifier"
```
Fefaine: I think it could be a good idea to filter the content of the input, if the user adds > or < or " it might break the XML right?
I would add an extra scenario for that purpose.

cyberwolf: Added additional test.

PR #943, opened 2016-10-22
switch from PhantomJS to Selenium2

fdewinne:
Hi,

We just installed separate containers into continuousphp dedicated to selenium to improve headless web driver stability.
This is an official container from selenium and you can easily use it locally for development purpose.
Just check it out from [Docker Hub](https://hub.docker.com/r/selenium/standalone-chrome/)

Into [continuousphp](https://continuousphp.com), this selenium container is accessible through http://selenium:4444/wd/hub url and the application through http://application. More services will be run from separate containers in the near future.

Best Regards,
Frederic


jfhovinne:
Thanks for the PR.

Pulled that branch + merged #936 (which fixes tests that did not pass any more) => success (on a local VM); all tests pass.

#936 still needs to be QAed though; so increasing its priority.

The readme (https://github.com/ec-europa/platform-dev/blob/master/README.md) will need to be updated to reflect the changes.

jfhovinne:
See also #944.

fdewinne:
I just fixed the place holders for the selenium params and injected the right values in continuousphp and the result is now ok. Keep in mind to inject these values on other pipelines if needed (behat.seleniumuri & behat.browser) in the first step. You should find the result as output message when the setup-behat phing target runs.

pfrenssen:
This is now fixed upstream in Mink: https://github.com/minkphp/MinkSelenium2Driver/pull/252

I will integrate that fix here. I will also rename the build properties so they are identical to the ones used in the CI fork of `drupal-project`, this will be more familiar for developers moving forward to Drupal 8 in the near future.

PR #964, opened 2016-10-27
NEPT-161: Upgrade menu_token to beta7 and remove patch

drupol:
fix: Upgrade menu_token to beta7.


drupol:
After checking differences between beta5 and beta7... I think this could be included in master too... but to be discussed.

cc @jfhovinne @Fefaine

PR #966, opened 2016-10-27
feat(interoperability): add cache purging feature set

cyberwolf:
Implements NEPT-79.


profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- 'nexteuropa_varnish_http_targets',
+- 'nexteuropa_varnish_tag',
+- 'nexteuropa_varnish_request_method'
+
+In order to enable the feature make sure that above variables are set
+and if so then go to the `admin/structure/feature-set` page select
+'Rule-based web frontend cache purging' and click 'Validate' button.
+
+## Custom entity - 'Purge rule'
+The Nexteuropa Varnish provides additional custom entity type:
```
Fefaine: You should start this section by explaining that the modules adds an additionnal configuration screen under 'Configuration' menu item.

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- Purge rule - machine name: `nexteuropa_varnish_cache_purge_rule`
+
+It allows to create and maintain a set of rules which are responsible
+for sending customized HTTP requests to the Varnish server and by
+that invalidating the frontend cache.
+
+To add and maintain purge rules go to the following url:
+`admin/config/frontend_cache_purge_rules`
```
Fefaine: It's better to direct the users to the config screen by indicating under which menu item it lies

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Purge rules logic
+The Nexteuropa Varnish provides hardcoded logic for triggering
+configured rules. Current version implements two workflow cases for:
+- content types moderated via the workbench moderation module
+- content types without additional moderation (default Drupal settings)
+
+### Content moderated via the workbench moderation module
+For the content types which are controlled via workbench moderation
+module, created purge rules would be triggered in the following cases:
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- always when the given content would have a workflow state change to 'Published'
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- always when the given content would have a workflow state change from 'Published' to the other
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Content without moderation
+For the content types which are not moderated (default Drupal content
+type with two states published and unpublished), created purge rules
+would be triggered in the following cases:
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- when the node of the given content type would be created and saved with the 'Publish' state
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,83 @@
...
+- when the published node of the given content type would be updated
```
Fefaine: use "will" instead of "would"

profiles/common/modules/custom/nexteuropa_varnish/README.md
Fefaine: You should provide an example of how the new rules can be added:
-How to select the variables
-Some basic example of patterns usage.

profiles/common/modules/custom/nexteuropa_varnish/README.md
Fefaine: You should mention that the module provides a new permissions which allows to create new rules.

profiles/common/modules/custom/nexteuropa_varnish/README.md
Fefaine: It is not clear if I select "Basic page" and add a path to an article in "A specific list of paths" what will happen.
I think the "Content type" select box should not apply the the "list of paths" selection

netlooker: It is explained. So you need to pick content type and based on that choose if you want to specify paths which should be purged or just purged the path which is related to the currently processed node.
If you pick 'Paths of the node the action is performed on' then you don't have possibility to enter paths.

netlooker: There is a separate User Manual on internal Confluence. I thought that in README.md we should put more technical details rather than user manuals.

profiles/common/modules/custom/multisite_drupal_features_set_standard/multisite_drupal_features_set_standard.info
```diff
@@ -50,6 +50,14 @@ feature_sets[nexteuropa_integration][font] = 'multisite-icon-share'
 feature_sets[nexteuropa_integration][enable][] = nexteuropa_integration
 feature_sets[nexteuropa_integration][disable][] = nexteuropa_integration
 
+feature_sets[nexteuropa_varnish][title] = 'Rule-based web frontend cache purging'
+feature_sets[nexteuropa_varnish][description] = 'Define rules to trigger granular purging of the web frontend caches'
+feature_sets[nexteuropa_varnish][documentation] = ''
```
voidtek: The link of the README.md is missing.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.install
```diff
@@ -64,6 +96,21 @@ function nexteuropa_varnish_install() {
...
+ */
+function nexteuropa_varnish_uninstall() {
+  $vars = [
+    'nexteuropa_varnish_http_targets',
+    'nexteuropa_varnish_tag',
+    'nexteuropa_varnish_request_method',
+  ];
+
+  foreach ($vars as $var) {
+    variable_del($var);
```
voidtek: We delete variables that we have not initiate it during the hook_install.
I think it's not necessary to delete it.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -273,83 +265,62 @@ function _nexteuropa_varnish_purge_paths($paths) {
...
-      array_keys($request['headers']),
-      $request['headers']
+    $url = sprintf('http://%s%s', $target, $request_path);
+
+    $response = drupal_http_request(
+      $url,
+      array(
+        'headers' => $request_headers,
+        'method' => $request_method,
+        'timeout' => variable_get('nexteuropa_varnish_http_timeout', 2.0),
```
voidtek: the variable "nexteuropa_varnish_http_timeout" is not set before this call or is not documented for the additional settings.

profiles/common/modules/custom/multisite_drupal_features_set_standard/multisite_drupal_features_set_standard.info
netlooker: I've added a link to the internal wiki page as it is more complete and user oriented.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
netlooker: I've updated docs.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -273,83 +265,62 @@ function _nexteuropa_varnish_purge_paths($paths) {
...
-    }
-
-    // Format headers.
-    $formatted_headers = array_map(
-      function ($name, $value) {
-        return sprintf('%s: %s', $name, trim($value));
-      },
-      array_keys($request['headers']),
-      $request['headers']
+    $url = sprintf('http://%s%s', $target, $request_path);
```
netlooker: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/README.md
```diff
@@ -0,0 +1,111 @@
...
+This feature can be enabled only with the support of the QA/Maintenance
+team.
+The following environment specific variables have to be configured
+before enabling the feature:
+- 'nexteuropa_varnish_request_user' - a string with the username
+- 'nexteuropa_varnish_request_password' - a string with the password
+- 'nexteuropa_varnish_http_targets', - an array with the urls ex. 'http://localhost'
+- 'nexteuropa_varnish_tag', - a string with the tag
+- 'nexteuropa_varnish_request_method' - a string with the HTTP request method
+- 'nexteuropa_varnish_http_timeout' - a float representing the maximum number
```
imanoleguskiza: The documentation specifies that the variable needs to be configured but then it is not checked in the hook requirements.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -83,8 +83,9 @@ function nexteuropa_varnish_cache_purge_rule_label($purge_rule) {
  */
 function nexteuropa_varnish_workbench_moderation_transition($node, $previous_state, $new_state) {
   $purge
-    = ($new_state == workbench_moderation_state_published()) ||
-      ($previous_state == workbench_moderation_state_published() && $new_state !== $previous_state);
+    = ($node->workbench_moderation['my_revision']->published) ||
```
imanoleguskiza: I don't understand why we need to check the workbench_moderation variable inside the node, are not the variables provided in the hook enough?
If I'm not wrong, the only moment where a PURGE request needs to be sent is when the $new_state is published (this is the moment when the publicly available content is changed)

profiles/common/modules/custom/nexteuropa_varnish/README.md
netlooker: I've added the variable `nexteuropa_varnish_http_timeout` to the hook_requirements implementation.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
netlooker: The PURGE request needs to be send in the following cases:
- change a moderation state to 'Publish' - ex. creation of a new page
- change a moderation state from 'Publish' to any other state - ex. unpublishing a page

voidtek:
:+1:

imanoleguskiza:
👍

voidtek:
:+1:

imanoleguskiza:
👍

PR #967, opened 2016-10-28
fix(contributing) Restore cache on full html

Fefaine:


Fefaine:
Won't do. See releated ticket for explainations

PR #981, opened 2016-10-31
Bugfix/nept 282

gillesdeudon:


Fefaine:
I am closing this PR. The right way to proceed for patches to add into 2.2 and 2.3 is to add the fix in master and merge master back to 2.3
Therefore this PR is useless

PR #986, opened 2016-11-03
NEXTEUROPA-14012: Fix \Drupal\user\Config::assignRoleToUser - multisite_config

vever001:
…RoleToUser & \Drupal\user\Config::revokeRoleFromUser


vever001:
I merged the master branch to solve the conflict with #1085 which I think is wrong because of 2 things:
1. It doesn't reset the static cache on user entities after adding/removing roles via a DB query.
2. It doesn't check for an ACTIVE og_membership state.

Fefaine:
Replaced with https://github.com/ec-europa/platform-dev/pull/1352

PR #992, opened 2016-11-07
NEPT-255: Missing default images during install.

drupol:
Fix: Missing user_default.png and empty_gallery.png files

This is to prevent warnings during install.


profiles/common/modules/features/communities/communities.install
```diff
@@ -31,16 +31,39 @@ function communities_enable() {
...
-  $f = file_save($file);
-  $_communities_fid_default_community_image = $f->fid;
+
+  $path = drupal_get_path('module', 'cce_basic_config') . '/assets/images/default_images';
+
+  $default_images = array(
+    'user_picture_default' => array(
+      'scheme' => 'public://',
+      'dest_directory' => 'default',
+      'source_path' => $path . '/community_default.png',
```
cyberwolf: Why is the source image inside cce_basic_config, instead of inside the communities feature?

profiles/common/modules/features/communities/communities.install
```diff
@@ -31,16 +31,39 @@ function communities_enable() {
...
+      'filemime' => file_get_mimetype($default_image_data['source_path']),
+    );
+    $file = file_save($file);
+
+    if ($file) {
+      variable_set($default_image_data['variable'], $file->fid);
+      $_communities_fid_default_community_image = $file->fid;
+    }
+  }
+
```
cyberwolf: It seems like communities_install() and _cce_basic_config_post_install_default_images() share some common code, which is now duplicated. Can this code be deduplicated into a helper method?

drupol: I though about it before doing it. But as things turns out with community profile, I don't know if it worth creating it this way.
Once that profile will not be used anymore we will have to redo the work and/or have extra code to maintain.

profiles/common/modules/features/communities/communities.install
drupol: oops this is a typo, I will fix and move the file.

drupol: Done.

profiles/common/modules/features/communities/communities.install
cyberwolf: I don't see how we would end up with extra code to maintain. The communities feature already has a dependency on cce_basic_config, so if we add the helper method with all common code to cce_basic_config we can call it both from the communities feature as well as from _cce_basic_config_post_install_default_images(). Less code to maintain.

Fefaine: We need the greenlight from @champcy and @gervasek to modify the community feature.

gervasek: Change on community feature are good for me, but yes I thinks we should avoid to duplicate code.
Like proposed by cyberwolf, we can create a helper function to import default images. For me, this helper function should be on "multisite_config" module.

The community profile is no longer used for new sites, but we need to keep this profile up to date, for maintenance on existing communities websites.

drupol:
Changes are up.

voidtek:
:+1:

profiles/common/modules/features/cce_basic_config/cce_basic_config.helpers.inc
```diff
@@ -0,0 +1,42 @@
...
+ *     - dest_directory: The destination directory
+ *     - source_path: The source path
+ *     - filename: The filename
+ *     - variable: The variable name containing the file id (fid) of the saved
+ *                 file if it has been successfully saved.
+ */
+function cce_basic_config_import_images($images) {
+  foreach ($images as $default_image_data) {
+    $file_uri = $default_image_data['scheme'] . $default_image_data['dest_directory'] . '/' . $default_image_data['filename'];
+    $full_dir = $default_image_data['scheme'] . $default_image_data['dest_directory'];
```
drupol: Use $full_dir in $file_dir

profiles/common/modules/features/communities/communities.install
gillesdeudon: :+1:

gillesdeudon:
:+1:

profiles/common/modules/features/communities/communities.install
dxvargas: 👍

PR #993, opened 2016-11-07
NEPT-161: Upgrade menu_token to beta7 and remove patch.

drupol:
fix: update to latest stable release, released in March 2016.


cyberwolf:
👍

verbruggenalex:
:-1: 
QA finished, we can not update the module until this issue is resolved:
https://www.drupal.org/node/2826475

entity_get_info does not always return path, for nodes it should still rely on entity_uri() to get the path

voidtek:
:+1:

drupol:
https://www.drupal.org/node/2826475 has been committed, there's only [one issue](https://www.drupal.org/node/2827674) but it's not a blocker.

drupol:
@verbruggenalex : as you're involved in menu_token, could you give us your green light for this ?

verbruggenalex:
:+1: I was hoping the maintainer would release a beta8 cause using a commit looks a bit ugly :) But I'm fine with it. Because I was partly responsible for writing the code I would like to have a second thumbs, @tatarbj  can you have a look at this?

Fefaine:
:+1:

Fefaine:
I am ok with using a commit, we do that quite often, even if we prefer avoiding it, sometimes it's less ugly than using the previous versions + 20 patches :D

netlooker:
:+1:

netlooker:
👍

Fefaine:
:+1:

PR #984, opened 2016-11-03
NEPT-261 Release/communities

Fefaine:

PR #1003, opened 2016-11-09
Release/communities delivery

gervasek:


tatarbj:
seems ok, but because it changes a lot of things in db, i don't think it's possible to be deployed as a hotfix or silent update - the code with the remarks by #984 is checked and accepted by me, the thumbs up is coming:

tatarbj:
:+1:

PR #1004, opened 2016-11-10
fix(contributing): Nexteuropa_editorial breaks display suite. NEPT-291

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-291


imanoleguskiza:
👍

tests/src/Context/DrupalContext.php
```diff
@@ -97,4 +104,44 @@ public function resetNodes() {
...
+   * Removes any node types created after the last list node type remembered.
+   *
+   * @AfterScenario @reset-node-types
+   */
+  public function resetNodeTypes() {
+    if (!isset($this->nodeTypes)) {
+      return;
+    }
+
+    $results = db_select('node_type')
```
netlooker: Similar like in the previous comment. Usage of the `node_type_get_names()` function would make it more consistent especially that below there is the `node_type_delete()` which is coming from the same functions family. In future that approach would be safer because it relies on the API rather than on the database queries.

tests/src/Context/MinkContext.php
```diff
@@ -309,6 +309,31 @@ public function getTableRow(Element $element, $search) {
...
+   * @Then I check if :div1 and :div2 have the same position from top
+   */
+  public function checkIfTwoDivHaveSamePosition($div1, $div2) {
+    $javascript1 = "return jQuery('." . $div1 . "').offset().top;";
+    $result_div1 = intval($this->getSession()->evaluateScript($javascript1));
+
+    $javascript2 = "return jQuery('." . $div2 . "').offset().top;";
+    $result_div2 = intval($this->getSession()->evaluateScript($javascript2));
+
+    if ($result_div1 != $result_div2) {
```
netlooker: Here I would prefer to use 'aasert' function as follow:
`assert($result_div1, equals($result_div2));`
It will simplify the step implementation to make it more readable and easier to maintain in future. This will also make the step more consistent with the recently developed steps implementations.

tests/src/Context/DrupalContext.php
```diff
@@ -97,4 +104,44 @@ public function resetNodes() {
     unset($this->maxNodeId);
   }
 
+  /**
+   * Remember the list of node type.
+   *
+   * @BeforeScenario @reset-node-types
+   */
+  public function rememberCurrentNodeTypes() {
+    $results = db_select('node_type')
```
voidtek: Good remark.

tests/src/Context/DrupalContext.php
voidtek: Good remark.

tests/src/Context/MinkContext.php
voidtek: Very good improvement, thanks.

imanoleguskiza:
👍

gillesdeudon:
:+1:

PR #1011, opened 2016-11-14
Feature/nept 313

jfhovinne:


Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    stage('Build') {
+        withCredentials([
+            [$class: 'UsernamePasswordMultiBinding', credentialsId: 'mysql', usernameVariable: 'DB_USER', passwordVariable: 'DB_PASS'],
+            [$class: 'UsernamePasswordMultiBinding', credentialsId: 'flickr', usernameVariable: 'FLICKR_KEY', passwordVariable: 'FLICKR_SECRET']
+        ]) {
+            sh 'mysql -u $DB_USER --password=$DB_PASS -e "CREATE database $DB_NAME;"'
+            sh 'composer install --no-suggest'
+            sh "./bin/phing build-platform-dev -Dcomposer.bin=`which composer` -D'behat.base_url'='$BASE_URL/$SITE_PATH/build' -D'env.FLICKR_KEY'='$FLICKR_KEY' -D'env.FLICKR_SECRET'='$FLICKR_SECRET' -D'integration.server.port'='$HTTP_MOCK_PORT' -D'varnish.server.port'='$HTTP_MOCK_PORT'"
```
drupol: Instead of specifying variables on the command line, why don't we use a build.properties file for jenkins ?

Jenkinsfile
```diff
@@ -0,0 +1,48 @@
...
+        deleteDir()
+        checkout scm
+    }
+
+    stage('Build') {
+        withCredentials([
+            [$class: 'UsernamePasswordMultiBinding', credentialsId: 'mysql', usernameVariable: 'DB_USER', passwordVariable: 'DB_PASS'],
+            [$class: 'UsernamePasswordMultiBinding', credentialsId: 'flickr', usernameVariable: 'FLICKR_KEY', passwordVariable: 'FLICKR_SECRET']
+        ]) {
+            sh 'mysql -u $DB_USER --password=$DB_PASS -e "CREATE database $DB_NAME;"'
```
drupol: I think this should be handled properly in a phing task and instead of calling shell commands, we should only use phing commands.

drupol: Think can be refactored in the next release.

Jenkinsfile
drupol: Think can be refactored in the next release.

drupol:
:+1:

PR #1018, opened 2016-11-16
fix(editorial): Add patch 2541786 for ds.

voidtek:
https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-330

### DIFF
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -170,6 +170,10 @@ projects[diff][version] = "3.2"
 
 projects[ds][subdir] = "contrib"
 projects[ds][version] = "2.11"
+; Issue #2541786 : Changing layout to "fluid 2 columns stacked" resulted in crash
+; https://www.drupal.org/node/2541786
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-330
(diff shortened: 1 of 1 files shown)


resources/multisite_drupal_standard.make
```diff
@@ -170,6 +170,10 @@ projects[diff][version] = "3.2"
 
 projects[ds][subdir] = "contrib"
 projects[ds][version] = "2.11"
+; Issue #2541786 : Changing layout to "fluid 2 columns stacked" resulted in crash
+; https://www.drupal.org/node/2541786
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-330
+projects[ds][patch][] = https://www.drupal.org/files/issues/2541786-8.patch
```
Fefaine: I disagree to solve this by patching 2.11
This patch is already included in 2.12
We should instead upgrade DS to 2.14
This is why It's important to include as much as possible module upgrades in 2.3 :)

Fefaine:
:-1:

resources/multisite_drupal_standard.make
voidtek: Ok.

Fefaine:
To be replaced by DS module upgrade

PR #1027, opened 2016-11-18
fix(contributing): Removing permission 'use media wysiwyg'

netlooker:
- removing 'use media wysiwyg' permission from the multisite_wysiwyg feature
- fixing failing test scenario (new step implementation)


imanoleguskiza:
👍

Fefaine:
I agree with @voidtek 
We should provide a hook_install that will grant this permission to administrator, editor and contributor
On a fresh install a contributor should be able to use the wysiwyg, that a basic editorial feature !

dxvargas:
👍

imanoleguskiza:
👍

PR #1034, opened 2016-11-22
Fix bugs related to nexteuropa editorial

bianchim:
NEXTEUROPA-14012

### DIFF
--- a/profiles/common/modules/features/nexteuropa_editorial/src/Config.php
+++ b/profiles/common/modules/features/nexteuropa_editorial/src/Config.php
@@ -21,14 +21,23 @@ class Config extends ConfigBase {
    *
    * @param int $uid
    *    User UID.
+   * @param int $new_group
+   *    The group we're adding to the user.
    *
    * @return bool
    *    TRUE if the user is a team member, FALSE otherwise.
    */
-  public function isEditoralTeamMember($uid = 0) {
+  public function isEditoralTeamMember($uid = 0, $new_group = NULL) {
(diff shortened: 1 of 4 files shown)


profiles/common/modules/features/nexteuropa_editorial/src/Config.php
```diff
@@ -21,14 +21,23 @@ class Config extends ConfigBase {
...
    * @return bool
    *    TRUE if the user is a team member, FALSE otherwise.
    */
-  public function isEditoralTeamMember($uid = 0) {
+  public function isEditoralTeamMember($uid = 0, $new_group = NULL) {
     $is_team_member = FALSE;
     $account = user_load($uid);
     $groups = og_get_groups_by_user($account, 'node');
+
+    // If in the process of adding a new group, add it since cache may not have
```
vever001: We are not inserting a new group here but a new og_membership.
IMO we shouldn't even use og_get_groups_by_user().

Actually this whole function can be replaced in 2 lines (taking OG membership states into account!):
$groups = og_get_entity_groups('user', $uid, array(OG_STATE_ACTIVE), 'og_user_node');
return !empty($groups['node']);

bianchim: We could implement a bit of code to reproduce this behavior but I don't see the point. The goal of this method is to determine if a user is active within a specific type of group, and we would lose these two conditions doing it that way.

PR #1006, opened 2016-11-10
NEPT-153: Rearrange behat tests parallelization

ademarco:

PR #1043, opened 2016-11-24
NEPT-367: Use proper projects for Composer dependencies (phing stuff)

drupol:
Phing Drush Task and Phing Behat Task are now projects on Drupal and Github respectively.

- https://github.com/drupol/phingdrushtask
- https://github.com/drupol/phingbehattask

Both projects has full tests coverage. It might be great to use those tools inside the platform.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,16 +1,16 @@
 {
   "require": {
+    "php": ">=5.4.0",
     "drupal/coder": "8.2.6",
-    "drupal/phing-drush-task": "1.0",
-    "drush/drush": ">=8.0.1",
+    "drupal/phingdrushtask": "^1.1",
+    "drupol/phingbehattask": "^1.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -1,16 +1,16 @@
 {
   "require": {
+    "php": ">=5.4.0",
```
ademarco: We should require PHP >= 5.6.0 since we use `use function` construct in our custom Behat contexts.

Fefaine: Also, our servers are runing 5.6 So i would even fix this to 5.6

drupol: This is out of scope and part of #1044

ademarco:
We decided with @drupol to include this in NEPT-153. Closing this PR as wont fix.

drupol:
Link #1006

PR #1044, opened 2016-11-24
NEPT-368: composer.json cleanup

drupol:
The platform is currently using 2 composer.json files.

We should put some order in there and update projects to their most up to date versions. (Coder)
Also, there is no require-dev section in them.

The require-dev packages are packages that aren't necessary for your project to work and that shouldn't be included in the production version of your project.
Typically, these are packages such as phpunit/phpunit that you would use for unit testing your project.


drupol:
JENKINS IS GREEN ! @ademarco ! YEAH !

ademarco:
Nice! But the pull request is too big to review it properly, could you please move the phpcs fixes commit on another branch and revert it on this one?

drupol:
PHPCS fix needs to be in a single PR.

PR #1046, opened 2016-11-25
Bugfix/nept 211

imanoleguskiza:


profiles/common/modules/custom/nexteuropa_token/src/Entity/ViewModeTokenHandler.php
```diff
@@ -154,7 +154,8 @@ protected function parseToken($original, $item = 'entity_id') {
   public function getEntityViewModes($token_type) {
 
     $view_modes = array();
-    $entity_type = $token_type == 'term' ? 'taxonomy_term' : $token_type;
+    $token_types = token_get_entity_mapping('token');
+    $entity_type = $token_types[$token_type];
```
netlooker: This can cause an error if $token_type is not present in the $token_types. Maybe isset() condition is worth to consider. It depends if the $token_type passed is always well known and there is no way to change it by the users actions. In function  getEntityTokenTypes() implementation is wrapped into condition check.

profiles/common/modules/custom/nexteuropa_token/src/Entity/UrlTokenHandler.php
```diff
@@ -47,10 +47,11 @@ public function hookTokens($type, $tokens, array $data = array(), array $options
     $replacements = array();
 
     if ($this->isValidTokenType($type)) {
+      $token_types = token_get_entity_mapping('token');
```
netlooker: Function `token_get_entity_mapping()` by default provides mapping for the 'token' value;

profiles/common/modules/custom/nexteuropa_token/src/Entity/ViewModeTokenHandler.php
```diff
@@ -154,7 +154,8 @@ protected function parseToken($original, $item = 'entity_id') {
   public function getEntityViewModes($token_type) {
 
     $view_modes = array();
-    $entity_type = $token_type == 'term' ? 'taxonomy_term' : $token_type;
+    $token_types = token_get_entity_mapping('token');
```
netlooker: Function `token_get_entity_mapping()` by default provides mapping for the 'token' value;

profiles/common/modules/custom/nexteuropa_token/src/Entity/LinkTokenHandler.php
```diff
@@ -47,11 +47,12 @@ public function hookTokens($type, $tokens, array $data = array(), array $options
     $replacements = array();
 
     if ($this->isValidTokenType($type)) {
+      $token_types = token_get_entity_mapping('token');
```
netlooker: Function token_get_entity_mapping() by default provides mapping for the 'token' value;

netlooker:
👍

dxvargas:
👍

PR #1060, opened 2016-11-29
Bugfix/nept 212

imanoleguskiza:


netlooker:
👍

profiles/common/modules/features/nexteuropa_core/nexteuropa_core.tokens.inc
```diff
@@ -27,13 +27,12 @@ function nexteuropa_core_tokens($type, $tokens, array $data = array(), array $op
...
       if (strstr($name, ':parents-uri') !== FALSE) {
-        list($field_name, $delta,) = explode(':', $name);
+        list($field_name) = explode(':', $name);
         $field_name = str_replace('-', '_', $field_name);
         if (isset($node->{$field_name})) {
           $node_wrapper = entity_metadata_wrapper('node', $node);
-          $terms = $node_wrapper->{$field_name}->value();
-          if (isset($terms[$delta - 1])) {
-            $term = $terms[$delta - 1];
+          $term = $node_wrapper->{$field_name}->value();
```
gillesdeudon: $node_wrapper->{$field_name}->value() can return an array if the field is multi-values (see last part of the ["Working with lists" section](https://www.drupal.org/docs/7/api/entity-api/entity-metadata-wrappers#example-value-set-save). Should we take it into account?

profiles/common/modules/features/nexteuropa_core/nexteuropa_core.tokens.inc
```diff
@@ -27,13 +27,12 @@ function nexteuropa_core_tokens($type, $tokens, array $data = array(), array $op
...
+          if (!empty($term)) {
             $parents = array();
             foreach (entity_metadata_taxonomy_term_get_properties($term, array(), 'parents_all') as $tid) {
```
gillesdeudon: Instead of "entity_metadata_taxonomy_term_get_properties", we should use taxonomy_get_parents_all. it retrieve all parent taxonomy terms what entity_metadata_taxonomy_term_get_properties and that could avoid to do another taxonomy_term_load in the loop.

profiles/common/modules/features/nexteuropa_core/nexteuropa_core.tokens.inc
imanoleguskiza: Not really, the token only applies to single term reference fields

imanoleguskiza: Hi Gilles,
I just made some tests and tokens already has validation to prevent the use of tokens in incorrect contexts. In this case, if someone tries to force this token with a multivalue term reference, he won't be able to. See token_element_validate inside the token module.
Based on this, I don't think we should implement extra validation steps in our code.

gillesdeudon: :+1:

gillesdeudon:
:+1:

voidtek:
:+1:

gillesdeudon:
:+1:

gillesdeudon:
I put directly the 2nd thumb up because the change was just a typo one.

PR #1072, opened 2016-12-02
NEPT-310: Patch plupload with a format valid git patch.

drupol:
Actually the drush make workflow automatically apply patches to projects.
In some cases, it wont work properly due to some version of the git and patch softwares.
This patch fix the problem by using the right standard patch that works everywhere.

Please read more information here: https://www.drupal.org/node/1903850

This is a blocker and this should go to master as the changes in the code are not impacting anything.

FYI, [here a link that](http://pastebin.com/wWLLXUfg) shows the diff between the version currently used (1.7) and the version I would like to use.

Please, could you think about it twice?


jfhovinne:
:+1:

gillesdeudon:
:+1:

PR #1087, opened 2016-12-08
Feature/nept 380

jfhovinne:


build.properties.dist
```diff
@@ -110,6 +110,12 @@ behat.yml.path = ${behat.dir}/behat.yml
 # The base URL to use in Behat tests.
 behat.base_url = http://localhost
 
+# The URL of the Behat webdriver host.
+behat.wd_host.url = http://localhost:8643/wd/hub
```
drupol: Why not use 4444 ? Everyone uses that on the internet, why such a change ? I understand it's not a big deal, I'm just trying to understand :)

jfhovinne: Legacy perhaps?
Anyway, it can be overridden.

resources/composer.json
```diff
@@ -3,8 +3,9 @@
     "guzzlehttp/guzzle": "^6.2.1"
   },
   "require-dev": {
-    "behat/behat": "~3.1.0@rc",
-    "drupal/drupal-extension": "~3.1.0",
+    "behat/behat": "~3.1",
+    "drupal/drupal-extension": "~3.1",
```
drupol: Use "drupal/drupal-extension": "dev-master#4d072006873a15a704764985c22dfb591c129c7e"

jfhovinne: Generates lots of notices which make the build to fail.
To be handled on another PR.

README.md
```diff
@@ -100,12 +100,20 @@ $ ./bin/phing setup-behat
 In order to run JavaScript in your Behat tests, you must launch a PhantomJS
 instance before. Use the `--debug` parameter to get more information. Please
 be sure that the webdriver's port you specify corresponds to the one in your
-Behat configuration (`wd_host: "http://localhost:8643/wd/hub"`).
+Behat configuration (`behat.wd_host.url: "http://localhost:8643/wd/hub"`).
```
voidtek: It's in the "build.properties.dist" with the sentence "behat.wd_host.url = http://localhost:8643/wd/hub". (and not in the Behat configuration - behat.xxx.yml.dist )
Maybe, the last sentence is a little confusing.

voidtek:
:+1:

build.properties.dist
Fefaine: That's by default in our VPS install. Maybe you could raise this to Ivan for the next VPS version?

Fefaine:
The tests that use javascript are failing on jenkins1, therefore i am not in favor of plussing this :fearful:

Jenkinsfile
```diff
@@ -33,14 +37,19 @@ node('master') {
 
         stage('Test') {
             wrap([$class: 'AnsiColorBuildWrapper', colorMapName: 'xterm']) {
-                sh './bin/behat -c build/behat.api.yml --colors -f pretty --strict'
-                sh './bin/behat -c build/behat.i18n.yml --colors -f pretty --strict'
+                timeout(time: 2, unit: 'HOURS') {
+                    if (env.WD_BROWSER_NAME == 'phantomjs') {
```
gillesdeudon: Ok. I discussed with @jfhovinne and @voidtek about this comment. The case I was talking about will only happen if the server is not well configured and then, it will be treated at the environment installation.

So, that can be considerate as a nice to have that can be treated in another ticket.

gillesdeudon:
:+1:

Jenkinsfile
jfhovinne: Thanks for the comments.
Yes, this will be environment-related then.
Same reply regarding failures on Jenkins 1.

PR #1088, opened 2016-12-09
NEPT-296: Upgrade workbench moderation to latest version - Security risk: 11/25 ( Moderately Critical)

ademarco:


resources/multisite_drupal_standard.make
```diff
@@ -746,28 +748,22 @@ projects[workbench_email][patch][] = https://www.drupal.org/files/issues/workben
...
-; https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-13039
-projects[workbench_moderation][patch][] = https://www.drupal.org/files/issues/node-deleted-before-shutdown-function-2645622-4.patch
-
 
 projects[workbench_og][subdir] = "contrib"
 projects[workbench_og][version] = "2.0-beta1"
+; Content not accessible after being published - node_access not updated
+; Issue https://www.drupal.org/node/2835937
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-296
+projects[workbench_og][patch][] = https://www.drupal.org/files/issues/workbench_og-node_access-2835937.patch
```
voidtek: The last patch for this issue in d.o is "workbench_og-node_access-2835937-8.patch"

dxvargas: When I wrote this there was only one patch.
Both patches do the same. I prefer this one and Gilles too. The patch is not so beautiful but the final code is much better. Anyway, being essentially the same, I don't think we need to change.

voidtek:
:+1:

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -82,12 +82,13 @@ function nexteuropa_varnish_cache_purge_rule_label($purge_rule) {
  * Implements hook_workbench_moderation_transition().
  */
 function nexteuropa_varnish_workbench_moderation_transition($node, $previous_state, $new_state) {
-  $purge
-    = ($new_state == workbench_moderation_state_published()) ||
-      (property_exists($node->workbench_moderation['my_revision'], 'unpublishing')
-        && $node->workbench_moderation['my_revision']->unpublishing);
+  $publishing = (($new_state == workbench_moderation_state_published()) &&
```
gillesdeudon: @voidtek, is it normal that when I have a published revision of a node and I only save a Draft revision of it, a "clearing" request is sent to Varnish?
I expected that in this case, no request is sent.

voidtek: @gillesdeudon, I confirm that this new case appears with this patch. I will fix it and create a new behat test with this scenario. Thanks.

gillesdeudon:
:+1:

voidtek:
:+1:

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
```diff
@@ -1303,18 +1303,6 @@ function multisite_drupal_toolbox_linkit_scan_url($url, $field) {
...
 /**
- * Implements hook_i18n_sync_translation().
- *
- * Fix bug: synchronise translations causes an infinite loop.
- *
- * @see https://webgate.ec.europa.eu/CITnet/jira/browse/MULTISITE-1636
- * @see http://drupal.org/node/1372500
- */
-function multisite_drupal_toolbox_i18n_sync_translation($entity_type, $translation, $translation_language, $source, $source_language, $field_names) {
-  $translation->workbench_moderation['updating_live_revision'] = TRUE;
```
gillesdeudon: For the record, I did not succeed to reproduce the problem when I gave the second :+1:

gillesdeudon:
:+1:

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -82,12 +82,16 @@ function nexteuropa_varnish_cache_purge_rule_label($purge_rule) {
  * Implements hook_workbench_moderation_transition().
  */
 function nexteuropa_varnish_workbench_moderation_transition($node, $previous_state, $new_state) {
-  $purge
-    = ($new_state == workbench_moderation_state_published()) ||
-      (property_exists($node->workbench_moderation['my_revision'], 'unpublishing')
-        && $node->workbench_moderation['my_revision']->unpublishing);
+  if (isset($node->is_draft_revision) ? $node->is_draft_revision : FALSE) {
```
drupol: This is the first time I see that :)

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -82,12 +82,16 @@ function nexteuropa_varnish_cache_purge_rule_label($purge_rule) {
...
+    return;
+  }
+
+  $publishing = $new_state == workbench_moderation_state_published();
+
+  $unpublishing = (($new_state != workbench_moderation_state_published()) &&
```
drupol: We could replace `$new_state != workbench_moderation_state_published()` with `!$publishing`.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -681,4 +681,52 @@ public static function cleanCommunityEnvironment() {
...
+
+  /**
+   * Assert field language given field name, content type and content title.
+   *
+   * @Then I should only have :arg1 in :arg2 for :arg3 content with title :arg4
+   */
+  public function assertFieldLanguageForContentWithTitle($field_name, $language, $type, $title) {
+    $node = $this->getNodeByTitle($type, $title);
+
+    $query = db_select("field_data_{$field_name}", 'f')
```
gillesdeudon: In the workbench moderation upgrade context, this test is problematic because it works only in the case of a content that has no published revision. 
In case of a content saved as draft while a published revision exist, that tests the language against the published revision but not against the the revision that was just saved. 
So, it is not generic enough.

I recommend then to use use query that uses the "field_revision_{$field_name}" instead and use the current revision that goes with the node loading normally.

or rename the step to indicate that we test against the published revision or the draft revision when no published revision exists

tests/features/subscriptions.feature
```diff
@@ -55,6 +55,30 @@ Feature: Subscription
...
+    # Check "Basic page" option
+    And I check "edit-0-checkboxes-page-1"
+    And I press "Save"
+    When I am logged in as a user with the 'administrator' role
+    When I go to "node/add/page"
+    And I fill in "Title" with "Another page"
+    And I click "Publishing options"
+    And I select "Published" from "Moderation state"
+    When I press "Save"
+    And I am on "admin/config/system/cron_en"
```
gillesdeudon: I see an interest here to create a follow-up ticket for creating a step "And I wait for the Drupal cron" that execute the cron like a crontab or a `drush core-cron`. here, we simlulate a Poormanscron execution what is not exactly the same as on a production envrionement;

profiles/common/modules/features/multisite_notifications/multisite_notifications_core/multisite_notifications_core.module
```diff
@@ -12,17 +12,11 @@ include_once 'multisite_notifications_core.features.inc';
...
-
-  // Only send notification if the state of the content is published.
-  if (isset($event['node']->workbench_moderation_state_new)) {
-    // Content has a worbench moderation workflow.
-    if ($event['node']->workbench_moderation_state_new != workbench_moderation_state_published()) {
+  // Is workbench_moderation overriding the workflow?
+  if (isset($event['node']->workbench_moderation)) {
+    // Only process events after publishing
+    // See multisite_notifications_core_workbench_moderation_transition()
+    if (!isset($event['node']->flag_after_publishing)) {
```
dxvargas: I've changed the name as suggested.
I don't see a practical advantage of setting the parameter to FALSE. But if you want to change that please do it, no problem also.

profiles/common/modules/features/multisite_notifications/multisite_notifications_core/multisite_notifications_core.module
```diff
@@ -38,8 +32,11 @@ function multisite_notifications_core_subscriptions_queue_alter(&$event) {
  * Implements hook_workbench_moderation_transition().
  */
 function multisite_notifications_core_workbench_moderation_transition($node, $previous_state, $new_state) {
+  // Only process events after publishing
+  // See multisite_notifications_core_subscriptions_queue_alter()
   if ($new_state == workbench_moderation_state_published()) {
     $full_node = node_load($node->nid);
+    $full_node->flag_after_publishing = TRUE;
```
dxvargas: As stated before, I don't see an advantage, but you can do it if you think it improves, no problem.

tests/features/subscriptions.feature
dxvargas: Maybe but here it shoudn't matter how the cron is fired, so this fits the propose here.
To cover the cron scenarios, ok, a new ticket could be interesting.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
dxvargas: This change was done by @voidtek .
For me it makes sense because it is the Workbench Moderation way. Parameters like this "is_draft_revision", but also others like "is_new" or "workbench_moderation_state_new" are some times defined in the process, other times they don't exist. That is not necessarily a problem as long as we handle it correctly like we do here.

profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
dxvargas: 👍

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
gillesdeudon: I see your point. The implementation is so disturbing. There is certainly an improvement to do here.
but that works, then. I will not block this point anymore.

gillesdeudon:
:+1:

gillesdeudon:
:+1:

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
voidtek: I confirm the @dxvargas comment.
The logic in Workbench Moderation must be improved and after, we could improve this part of the code to be more readable.

voidtek:
:+1:

voidtek:
:+1:

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -82,12 +82,17 @@ function nexteuropa_varnish_cache_purge_rule_label($purge_rule) {
  * Implements hook_workbench_moderation_transition().
  */
 function nexteuropa_varnish_workbench_moderation_transition($node, $previous_state, $new_state) {
-  $purge
-    = ($new_state == workbench_moderation_state_published()) ||
-      (property_exists($node->workbench_moderation['my_revision'], 'unpublishing')
-        && $node->workbench_moderation['my_revision']->unpublishing);
+  $is_draft_revision = isset($node->is_draft_revision) ? $node->is_draft_revision : FALSE;
+  if ($is_draft_revision) {
```
drupol: Or `if (isset($node->is_draft_revision) && $node->is_draft_revision)`

voidtek: It's done and committed.

voidtek:
:+1:

gillesdeudon:
:+1:

PR #1091, opened 2016-12-12
NEPT-412: Fix the security issue in contrib module Title.

drupol:
This patch will fix the security issue and it will also fix https://www.drupal.org/node/2757739.

So check_plain() encodes special characters that has special meaning in HTML (such as < and &) into plain text entities (i.e. &lt; and &amp; respectively) that will make these be rendered literally (not interpreted as HTML) when that string that is then displayed as part of a page with HTML-markup. The function filter_xss() filters an HTML string to prevent cross-site-scripting (XSS) vulnerabilities.
It does four things:

1. Removing characters and constructs that can trick browsers
2. Making sure all HTML entities are well-formed
3. Making sure all HTML tags and attributes are well-formed
4. Making sure no HTML tags contain URLs with a disallowed protocol (e.g. javascript:)
5. Both functions are used to sanitize data from users to make sure that any user injection is neutralized before the data is rendered on your site.

If you use **check_plain()** then the string passed to the function is supposed to be used as plain text (not HTML). Then filter_xss() is not needed, since check_plain() will always make the string plain text.

If you use **filter_xss()**, then the string passed to the function is supposed to be HTML, and check_plain() will mess it up.

This is why, in this particular case, **check_plain()** is used in **title_field_text_sync_get()** and **filter_xss()** in **title_field_text_with_summary_sync_get()**.


verbruggenalex:
:+1: passed QA

tatarbj:
@drupol 
The test_with_summary field type is defined by text.module that clearly checks the output of it, so the second filter_xss is not necessary: https://api.drupal.org/api/drupal/modules%21field%21modules%21text%21text.module/function/_text_sanitize/7.x (Actually it also uses check_markup or check_plain depends on the text processing)
I've also checked where it is called from, and it's by the taxonomy descriptions, after functionally testing it without your provided patch, all of our text filter works and doesn't have the issue that is in the scope of this ticket (the checked filters: full_html, full_html_track, filter_html, plain_text, basic_html).
Please remove that fix from the patch.

drupol:
Hi,

While making the tests, @voidtek made a breakthrough with the patch I tried to introduce.
After debugging we came to the conclusion that our approach was not the best.

Here's the best way to fix this.

verbruggenalex:
:-1:  
When using a filter format on the title field and entering a value in plain text field without wysiwyg I receive the following errors:
- Warning: html_entity_decode() expects parameter 1 to be string, array given in decode_entities() (line 463 of /opt/rh/httpd24/root/var/www/html/NEPT-412/build/includes/unicode.inc).
- Warning: Illegal offset type in pathauto_cleanstring() (line 237 of /opt/rh/httpd24/root/var/www/html/NEPT-412/profiles/multisite_drupal_standard/modules/contrib/pathauto/pathauto.inc).
- Notice: Array to string conversion in token_replace() (line 102 of /opt/rh/httpd24/root/var/www/html/NEPT-412/build/includes/token.inc).

To reproduce, do the following:
- edit replaced title field
- change text processing to filtered text
- add new content
- click "disable rich text" link (cause wysiwyg filters the script anyway I think)
- enter "<script>alert("Hello, World");</script>" as value
- save content

tests/features/content_editing_admin.feature
```diff
@@ -32,20 +31,19 @@ Feature: Content editing as administrator
...
-    | <div id=\"2invalidid\">A container with an invalid HTML ID</div>    | invalidid                     |
-    | <div id=\"invalidid.\">A container with an invalid HTML ID</div>    | invalidid                     |
+    Examples:
+      | title                                                                 | html                                                                  | unexpected                                                              |
+      | <script>alert('xss');</script>                                        | <script>alert('xss');</script>                                        | <script>alert('xss');</script>                                          |
+      | <a href=\"javascript:alert('xss');\">xss</a>                          | <a href=\"javascript:alert('xss');\">xss</a>                          | <a href=\"javascript:alert('xss');\">xss</a>                            |
+      | <p style=\"background-image: url(javascript:alert('xss');)\">xss</p>  | <p style=\"background-image: url(javascript:alert('xss');)\">xss</p>  | <p style=\"background-image: url(javascript:alert('xss');)\">xss</p>    |
```
drupol: ;

verbruggenalex:
Sorry for flip-flopping all over the place. But the core node title field does not allow for text processing.
The replaced field of the title module does. So my previous request to implement a similar approach as the core body/summary field, still stands:
https://api.drupal.org/api/drupal/modules!node!node.tokens.inc/function/node_tokens/7.x

gillesdeudon:
:+1:

dxvargas:
👍

gillesdeudon:
:+1:

PR #1097, opened 2016-12-13
Feature/nept 403

Fefaine:


profiles/common/modules/features/contact_form/contact_form.install
```diff
@@ -35,4 +35,42 @@ function contact_form_disable() {
   menu_link_delete(NULL, 'contact');
   // Activation message.
   drupal_set_message(t('Contact form feature is now disabled on your site.'));
+  module_disable(array('contact'), FALSE);
```
ademarco: We should not disable modules the current module depends upon as other modules might depend from  `contact`.

tests/features/contact_form.feature
```diff
@@ -0,0 +1,23 @@
...
+Feature: Contact Form
+  In order to allow sending emails to user without knowing their mail address
+  As a citizen of the European Union
+  I want to be able to access contact forms
+
+Scenario: Anonymous user can see the contact page
+  Given I am not logged in
+  And the module is enabled
+  | modules     |
+  | contact     |
```
ademarco: Shouldn't this be `contact_form`?

Fefaine: Yes! I'll fix it

profiles/common/modules/features/contact_form/contact_form.install
ademarco: Still not convinced about this sorry :( the issue says:

> Can the default alias be unset, or can the contact module be made optional.

So we should just do that, meaning removing it from `cce_basic_config` is all we need to do, then they can disable it manually on the info site since, I guess, `contact_form` module it's not enabled on their site.

Running `module_disable(array('contact'), FALSE);` is safe but it's still a strange behaviour that might rise questions down the road, I would just remove it, unless I'm missing something.

Fefaine: The issue was created by INFO site, so they focus on their problem. They will have to run a hook update to remove the module when moving to 2.3, Tamas is already informed.
We want to have a broader scope. and provide a 'on demand' service that enables contact form module through a feature.
We gradually move all bundled features to 'on demand' features (changed tracked, poetry and now contact)
If you disable poetry feature, poetry module is disabled. This is the same principle except that 'contact' is contrib.
If users want to use contact and not contact_form, they need to create their own feature with contact dependency.

ademarco:
👍

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+    And I fill in "Subject" with "Complaint"
+    And I fill in "Message" with "I am not happy with this contact page"
+    And I press the "Send message" button
+    Then I should see the following error messages:
+      | error messages                   |
+      | Math question field is required. |
+
+  @javascript
+  Scenario: Authenticated user can submit the contact page
+    Given I am logged in as a user with the administrator role
```
voidtek: The scenario should testes an "authenticated user" role and not an "administrator" role.

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
+@api
+Feature: Contact Form
+  In order to allow sending emails to user without knowing their mail address
+  As a citizen of the European Union
+  I want to be able to access contact forms
+
+  @javascript
+  Scenario: Anonymous user can see the contact page
```
voidtek: ... "and can submit a message."

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  @javascript
+  Scenario: Authenticated user can submit the contact page
+    Given I am logged in as a user with the administrator role
+    And the module is enabled
+      | modules      |
+      | contact_form |
+    When I am on "contact"
+    Then I should see "Contact - European Commission" in the "title" tag
+    When I fill in "Your name" with "Chuck"
```
voidtek: With an authenticated user, the name filed is already filled. (may be we can test it, optional)

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+    And I fill in "Your e-mail address" with "chuck.norris@improbabledommainname.com"
```
voidtek: With an authenticated user, the email address field is already filled. (may be we can test it, optional)

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+      | Math question field is required. |
+
+  @javascript
+  Scenario: Authenticated user can submit the contact page
+    Given I am logged in as a user with the administrator role
+    And the module is enabled
+      | modules      |
+      | contact_form |
+    When I am on "contact"
+    Then I should see "Contact - European Commission" in the "title" tag
```
voidtek: This check is not in the scope of this scenario.

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
```
voidtek: This check is not in the scope of this scenario.

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+    When I fill in "Your name" with "Chuck"
+    And I fill in "Your e-mail address" with "chuck.norris@improbabledommainname.com"
```
voidtek: With an authenticated user, the email address field is already filled. (may be we can test it, optional)

tests/src/Context/MediaContext.php
```diff
@@ -34,6 +34,20 @@ public function gatherContexts(BeforeScenarioScope $scope) {
   }
 
   /**
+   * Test the content of a unique tag with no css.
+   *
+   * @Then I should see :text in the :tag tag
+   */
+  public function iShouldSeeInTheTag($text, $tag) {
```
voidtek: I think that this function should be in the MinkContext.php

tests/features/contact_form.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  @javascript
+  Scenario: Authenticated user can submit the contact page
+    Given I am logged in as a user with the administrator role
+    And the module is enabled
+      | modules      |
+      | contact_form |
+    When I am on "contact"
+    Then I should see "Contact - European Commission" in the "title" tag
+    When I fill in "Your name" with "Chuck"
```
Fefaine: Yes but you can change it. We could add a follow up ticket to improve the test, is that ok for you?

tests/features/contact_form.feature
Fefaine: There was a test on this and I removed it when i removed the module from the profile, so i don't want to loose a test.
We need to discuss how to better define tests, but that is not in the scope. I just want to ensure all tests are kept.

tests/features/contact_form.feature
Fefaine: Let' create a follow up taks

tests/features/contact_form.feature
Fefaine: OK

gillesdeudon:
:+1:

Fefaine:
:+1:

dxvargas:
:+1:

PR #1113, opened 2016-12-19
fix(Contributing): Replace menu_get_object by path parsing - NEPT-416

gillesdeudon:


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -450,11 +450,25 @@ function nexteuropa_multilingual_path_update($path) {
...
-  // a translation mechanism different from entity translation.
-  $node = menu_get_object("node", 1, $path['source']);
-  if (isset($node->type)) {
-    $translation_mechanism = variable_get('language_content_type_' . $node->type, 0);
+  // We must pass through a basic path parsing and a db_select because:
+  // - The menu_get_object function makes a access control and a redirection to
+  // the home page if users do not have access permission.
+  // (see https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-416)
+  // - The logic is only applied to nodes that have "entity translation" as
+  // translation mechanism.
```
dxvargas: Should be "The logic is only applied to nodes that *don't* have "entity translation" as translation mechanism"

profiles/common/modules/custom/tmgmt_poetry/src/tmgmt_poetry.behat.inc
```diff
@@ -148,12 +148,14 @@ class TmgmtPoetrySubContext extends RawDrupalContext implements DrupalSubContext
 
 
   /**
-   * Get the Current Node Id in the URL.
+   * Get the Current Node Id.
    */
   private function getCurrentNodeId() {
-    $url = $this->getSession()->getCurrentUrl();
-    preg_match_all('/node\/(\d*)/', $url, $matches);
-    return (isset($matches[1][0]) ? $matches[1][0] : FALSE);
```
jfhovinne: Agreed. Should be for instance getLastCreatedNodeId().
Also, are we sure we will always get the id we expect (in this ticket context I mean) ?

profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -450,11 +450,25 @@ function nexteuropa_multilingual_path_update($path) {
...
+  $parsed_source = explode('/', $path['source']);
+  if (($parsed_source[0] != 'node')
+    || !isset($parsed_source[1])
+    || !is_numeric($parsed_source[1])) {
+    return;
+  }
+  $nid = $parsed_source[1];
+  $query = db_select('node', 'n')->condition('n.nid', $nid);
+  $query->addField('n', 'type', 'node_type');
+  $node_type = $query->execute()->fetchField();
```
gillesdeudon: I did it in the ticket.

profiles/common/modules/custom/tmgmt_poetry/src/tmgmt_poetry.behat.inc
gillesdeudon: Ok.

For your question, I say yes because in scenario using this method, they always implies one content saved just before and the last one is current one. Parallel processes implies different DB if my information is correct.

Now, while I am writing these lines, I concede that it is not CI agnostic . Then I will revert this commit. We will see for another solution in the future.

dxvargas:
👍

Fefaine:
:+1:

PR #1124, opened 2017-01-03
NEPT-430: analysis of the custom module multisite_drupal_toolbox...

drupol:
NEPT-430: analysis of the custom module multisite_drupal_toolbox, add missing dependencies in info file.

### DIFF
--- a/profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.info
+++ b/profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.info
@@ -4,5 +4,12 @@ package = multisite drupal toolbox
 core = 7.x
 
 dependencies[] = libraries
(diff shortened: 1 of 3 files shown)


profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.info
```diff
@@ -4,5 +4,12 @@ package = multisite drupal toolbox
 core = 7.x
 
 dependencies[] = libraries
+dependencies[] = multisite_config
+dependencies[] = linkchecker
+dependencies[] = entity
```
imanoleguskiza: multisite_config already has a dependency to entity so we should probably remove it.

imanoleguskiza:
👍

dxvargas:
👍

PR #1131, opened 2017-01-04
Update to the latest stable version of Behat

pfrenssen:
We are using Behat Drupal Extension for our tests, which up to now depended on an old release candidate of Behat (3.1.0-rc2 to be exact). The latest stable version of Behat is 3.3.0.

Let's see what happens if we run our tests on the latest version. There have been some important stability fixes in recent versions of Behat, but also some potential backwards compatibility breaking changes.

One example of a B/C break is that all callbacks like `@beforeNodeCreate` now need to be declared as static methods.


pfrenssen:
Note that this is not intended to be merged at this time, this is an exploratory PR to check if we will have to deal with potential compatibility breaks in the near future.

pfrenssen:
This looks very good. The tests are passing with a minor change to one of the step definitions. In the upcoming 4.x release of Behat Drupal Extension there is a `UserManager` service that can be used to retrieve the current user. I got the user from this service rather than working around it by snatching the user out of another context class.

As long as we stay on BDE 3.x this change won't even be necessary.

The exploration is a success. The platform is ready for Behat 3.3. This can be closed.

drupol:
Mmmh I like that, thx

PR #1132, opened 2017-01-05
Nept 225 : refactore field.multilingual.feature 

gillesdeudon:


tests/src/Context/DrupalContext.php
```diff
@@ -31,6 +31,13 @@ class DrupalContext extends DrupalExtensionDrupalContext {
   protected $nodeTypes;
 
   /**
+   * The list of field instances created during scenarios.
+   *
+   * @var array
+   */
+  protected $createdFieldList = array();
```
voidtek: The name of this variable is not good, in the code you use createFieldList.

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
+@api
+Feature: Content type administration features
+  In order to easily structure the content of the European Commission
+  As an administrator
+  I want to be able to create, customize and delete content types
+
+  Background:
+    Given I am logged in as a user with the 'administrator' role
+
+  Scenario: As administrator, I create a custom content type translated through
```
voidtek: All scenarios for this feature are for Administrator. I think it's not necessary to specify again here.

tests/src/Context/DrupalContext.php
```diff
@@ -209,4 +216,111 @@ public function theFollowingContentsUsingForWysiwygFields($text_format, TableNod
...
+
+        case 'Name':
+          $field_name = 'field_' . $value;
+          $this->createFieldList[$field_name] = array(
+            'field_name' => $field_name,
+            'entity_type' => 'node',
+            'bundle' => $arg1,
+          );
+          $is_field_exiting = field_info_field($field_name);
+          if ($is_field_exiting) {
```
voidtek: Why do you check the existing of this field after to add it in the createFieldList array.
I think we could test it before that.
And may be, it's better to use the function assert to check the existing of this field.

tests/src/Context/DrupalContext.php
```diff
@@ -209,4 +216,111 @@ public function theFollowingContentsUsingForWysiwygFields($text_format, TableNod
...
+    // Assign fields to user before creation.
+    foreach ($settings->getRowsHash() as $setting_name => $value) {
+      switch ($setting_name) {
+        case 'Label':
+          $field_instance_info['label'] = $value;
+          break;
+
+        case 'Name':
+          $field_name = 'field_' . $value;
+          $this->createFieldList[$field_name] = array(
```
voidtek: Maybe, we could add this field in the array only after the foreach. It's more simple to understand the logic and we limit errors by bad settings.

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    entity translation, using workbench moderation and having 2 fields "Body" and
+    "Select option". I should be able to create a content with it
+    When I go to "admin/structure/types/add"
+    And I fill in "Name" with "Groovy type"
+    And I fill in "Machine-readable name" with "groovy_type"
```
voidtek: This step is more technical than business.
We could remove this step and add the tag "@javascript" to this scenario.

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I fill in "Description" with "A Groovy type for a groovy test"
+    And I uncheck the box "Published"
+    And I uncheck the box "Promoted to front page"
+    But I check the box "Create new revision"
+    And I check the box "Enable moderation of revisions"
+    And I select the radio button "Enabled, with field translation" with the id "edit-language-content-type-4"
+    And I press the "Save and add fields" button for saving the "groovy_type" content type
+    Then I should be on "admin/structure/types/manage/groovy-type/fields_en"
```
voidtek: Could be modified by: 
Then I should see the success message "The content type Groovy type has been added."

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I should see the success message "The content type Groovy type has been added."
+    And I should see "Body"
+    When I fill in "edit-fields-add-new-field-label" with "Select an option"
+    And I fill in "edit-fields-add-new-field-field-name" with "select_an_option"
```
voidtek: This step is more technical than business.
We could remove this step and add the tag "@javascript" to this scenario.

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I select "list_text" from "edit-fields-add-new-field-type"
```
voidtek: We could use the business name of this option tag: "list_text" should be "List (text)"

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I select "options_buttons" from "edit-fields-add-new-field-widget-type"
```
voidtek: We could use the business name of this option tag: "options_buttons" should be "Check boxes/radio buttons"

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I press the "Save" button for saving the "field_select_an_option" field for the "groovy_type" content type
+    And I fill in "edit-field-settings-allowed-values" with:
+    """
+    Option 1
+    Option 2
+    """
+    And I press the "Save field settings" button
```
voidtek: After this step, we could add the following step to confirm the update:
Then I should see the success message "Updated field Select an option field settings."

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    And I check the box "edit-field-translatable"
+    And I press the "Save settings" button
+    Then I should see the success message "Saved Select an option configuration."
+    When I go to "node/add/groovy-type"
+    And I fill in "Title" with "Lorem ipsum dolor sit amet"
+    And I fill in "Body" with "<p>Consectetur adipiscing elit.</p>"
+    And I select the radio button "Option 1" with the id "edit-field-select-an-option-und-option-1"
+    And I press "Save"
```
voidtek: After this step, we could add the following step to check the creation: 
Then I should see the success message "Groovy type Lorem ipsum dolor sit amet has been created."

tests/features/content_types/content_type_administration.feature
```diff
@@ -0,0 +1,48 @@
...
+    Option 1
+    Option 2
+    """
+    And I press the "Save field settings" button
+    And I check the box "edit-field-translatable"
+    And I press the "Save settings" button
+    Then I should see the success message "Saved Select an option configuration."
+    When I go to "node/add/groovy-type"
+    And I fill in "Title" with "Lorem ipsum dolor sit amet"
+    And I fill in "Body" with "<p>Consectetur adipiscing elit.</p>"
```
voidtek: Why do you put the "<p>" tag in the body? 
The field Body uses the WYSIWYG editor.

tests/features/field.multilingual.feature
```diff
@@ -6,92 +6,78 @@ Feature: Field Multilingual features
...
-	And I press the "Delete" button
-	Then I should see the success message "The group My Group Color has been deleted from the Basic page content type."
+  Scenario: Administrator can translate field and field group labels. Once it is done,
+  users see field and field group labels translated in interface language or in the source language
+  if translations are not done
+    Given I am logged in as a user with the 'administrator' role
+    # Add and translate field and field Group
+    And a field with the following settings is added to "page" type:
+      | Label           | Select a Color  |
+      | Name            | selectcolor     |
```
voidtek: The name doesn't have the traditional syntax. In this case, it should be "field_select_a_color‎".

tests/features/field.multilingual.feature
```diff
@@ -6,92 +6,78 @@ Feature: Field Multilingual features
...
+    And I press the "Save translation" button
+    Then I should see the success message "3 translations were saved successfully."
+    # Field group creation: groupcolor.
+    When I go to "admin/structure/types/manage/page/display"
+    And I fill in "edit-fields-add-new-group-label" with "My Group Color"
+    And I fill in "edit-fields-add-new-group-group-name" with "groupcolor"
+    And I select "_add_new_group" from "edit-fields-field-selectcolor-parent"
+    And I press the "Save" button
+    Then I should see the success message "Your settings have been saved."
+    Then I should see the success message "New group My Group Color successfully created."
```
voidtek: "Then" should be modified by "And"

tests/features/field.multilingual.feature
```diff
@@ -6,92 +6,78 @@ Feature: Field Multilingual features
...
+    # Field group translation: groupcolor.
+    When I go to "content/title-english_en"
+    And I click "New draft"
+    And I select the radio button "Red" with the id "edit-field-selectcolor-en-red"
+    And I select "published" from "edit-workbench-moderation-state-new"
+    And I press the "Save" button
+    Then I should see the success message "Basic page Title in English has been updated."
+    When I go to "content/title-english_fr"
```
voidtek: This step is useless.

tests/features/content_types/content_type_administration.feature
gillesdeudon: I tried several ways to do it with @javascript and that failed all the time. Taking into account the time on it, I have chosen this way on purpose.

tests/features/content_types/content_type_administration.feature
gillesdeudon: I tried several ways to do it with @javascript and that failed all the time. Taking into account the time on it, I have chosen this way on purpose.

tests/features/content_types/content_type_administration.feature
gillesdeudon: I tried several ways to do it with @javascript and that failed all the time. Taking into account the time on it, I have chosen this way on purpose.

tests/features/content_types/content_type_administration.feature
gillesdeudon: it is a legacy from the previous test.

tests/features/content_types/content_type_administration.feature
gillesdeudon: Useless as we test the heading and the presence of the body and the other field.

tests/features/field.multilingual.feature
gillesdeudon: I kept the name set in the old scenario.

tests/features/field.multilingual.feature
gillesdeudon: :+1:

tests/src/Context/DrupalContext.php
gillesdeudon: :+1:

tests/src/Context/DrupalContext.php
gillesdeudon: :+1:

tests/src/Context/DrupalContext.php
gillesdeudon: The purpose of this code is purely technical. it is to avoid that someone creates a field definition that could change the configuration of an field of the platform. According to me, that will be useful while defining new scenarii that use this step.

tests/features/content_types/content_type_administration.feature
gillesdeudon: Legacy

tests/features/content_types/content_type_administration.feature
gillesdeudon: Legacy

tests/features/field.multilingual.feature
gillesdeudon: In fact, it is mandatory in order to allow the translation system registering "My group color" in strings to translate. 
Without this step, the rest of the scenario fails.

Fefaine:
This needs to be rebased to 2.4.0

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,203 @@
...
+    foreach ($settings->getRowsHash() as $setting_name => $value) {
+      switch ($setting_name) {
+        case 'Label':
+          $field_instance_info['label'] = $value;
+          break;
+
+        case 'Name':
+          $field_name = 'field_' . $value;
+          $is_field_exiting = field_info_field($field_name);
+          if ($is_field_exiting) {
```
voidtek: Could use the assert function here

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,203 @@
...
+      $this->defaultNodeTypes[$result->type] = $result->type;
+    }
+  }
+
+  /**
+   * Removes any node types created after the last list node type remembered.
+   *
+   * @AfterScenario
+   */
+  public function resetNodeTypes() {
```
voidtek: With this function, we can remove the following code (file FeatureContext.php - line 653)

    // Delete 'community' node type.
    _node_types_build(TRUE);
    node_type_delete('community');
    field_purge_batch(1);

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,203 @@
...
+   * Add a field group in the content type view and assign fields as children.
+   *
+   * @param string $arg1
+   *    The machine name of the content type to which attaching the field.
+   * @param TableNode $settings
+   *    Two columns table containing field settings.
+   *
+   * @Given a field group with the following settings is added to the :arg1 type view:
+   */
+  public function aFieldGroupWithTheFollowingSettingsIsAddedToTheTypeView($arg1, TableNode $settings) {
```
voidtek: This function doesn't check if the group :arg1 already exists.

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,203 @@
...
+    $group_machine_type = 'fieldset';
+    $mode = 'default';
+    $group_children = array();
+    // TODO : implement the fully support of the group type, missing
```
voidtek: Can you please create a follow-up for this TODO.

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    $current_fields = field_info_field_map();
+    $purge_field_schema = FALSE;
+    foreach ($current_fields as $field_name => $field_info) {
+      if (!isset($this->defaultFields[$field_name])) {
+        $field_info = field_info_field($field_name);
+        if (!is_null($field_info)
+          && !$field_info['locked']
+          && !$field_info['deleted']) {
```
voidtek: According to the documentation, the field_info_field only return a non-deleted fields.

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+          field_delete_field($field_name);
+          $purge_field_schema = TRUE;
+        }
+      }
+      else {
+        // We check if a test has not created an instance of an existing field.
```
voidtek: This part of the code is very interesting and very well implemented.
Unfortunately, I think, it's out of the scope of this ticket because I don't see any case on features with the case "create a instance between an existing field and an entity type"
But I think this logic to check field's changes, should be implemented on a follow up ticket. Furthermore, I think this code could be improved with other checks (for example: other values into the Field object like "type", "translatable", "locked", ... )

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+   * Removes any fields group created by a scenario.
+   *
+   * @AfterScenario
+   */
+  public function resetFieldGroups() {
+    if (empty($this->defaultFieldGroups)) {
+      return;
+    }
+    $current_field_groups = field_group_read_groups();
+    $this->scanFieldGroupsForResetting($current_field_groups, $this->defaultFieldGroups);
```
voidtek: "scanFieldGroupsForResetting" is a private function so I think that it's not necessary to pass "$this->defaultFieldGroups" in parameter.
You could remove "$this->defaultFieldGroups" in this call, and put it as the default value in the declaration of the function. (as you did with the parameter $checked_def_level)

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+
+        case 'bundle':
+          $this->scanFieldGroupsForResetting($item, $current_default_def_level, 'group_type');
+          break;
+
+        case 'group_type':
+          $this->scanFieldGroupsForResetting($item, $current_default_def_level, 'group');
+          break;
+
+        default:
```
voidtek: Could be "case 'group':"

tests/src/Context/FieldContext.php
gillesdeudon: Actually, we have a scenario where an instance of an existing field is added to article content type. see ECEmbeddedImageSubContext in ec_embedded_image.behat.inc

I implemented all elements necessary for existing scenarios and leave those that are not mandatory for them like complete the method that create a field group and that should be fixed during another ticket.

tests/src/Context/ContentTypeContext.php
gillesdeudon: I will not because that it is easier to work on this method while working on concrete cases, and adding a new element in it when it is necessary.

The todo comment is there for just reminding the method is fully implemented and we are aware  of it.

tests/src/Context/FieldContext.php
gillesdeudon: It is done on purpose.

tests/src/Context/FieldContext.php
gillesdeudon: I prefer my way to implement because it is more explicit, more flexible (even if it is a private method, it can be used in different cases inside the class in the future.)

tests/src/Context/FieldContext.php
gillesdeudon: I put a question on the doc because I had to implement that after meeting problems and the documentation does not say "field_info_field only return a non-deleted fields", it says "$field_name can only refer to a non-deleted, active field." and "NULL if the field was not found." that is not precisely the same.

tests/src/Context/DrupalContext.php
```diff
@@ -107,36 +100,6 @@ public function resetNodes() {
   }
 
   /**
-   * Remember the list of node type.
-   *
-   * @BeforeScenario @reset-node-types
```
gillesdeudon: 👍

tests/src/Context/ContentTypeContext.php
gillesdeudon: 👍

tests/src/Context/ContentTypeContext.php
gillesdeudon: My previous comment on this kind of remarks is hidden because of new commits.

I do not use assert here because this step is just a preparatory step for the actual test comping just after. 
In this case, this test is justified for simplifying the definition of the test during the development phase; it is not intended to be a validation step.
Then, the use of an exception is more appropriate.

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
```diff
@@ -86,34 +86,19 @@ class ECEmbeddedImageSubContext extends DrupalSubContextBase {
...
    * @AfterScenario
    */
-  public function removeMediaGalleryContentType() {
-    if (!$this->contentTypeWasSetUp) {
+  public function resetMediaGalleryPermission() {
+    if (!$this->permissionWasSetUp) {
       return;
     }
 
-    $this->contentTypeWasSetUp = FALSE;
```
voidtek: Should does not be removed.

tests/src/Context/ContentTypeContext.php
```diff
@@ -81,7 +81,12 @@ public function aFieldWithTheFollowingSettingsIsAddedToTheType($arg1, TableNode
           $field_name = 'field_' . $value;
           $is_field_exiting = field_info_field($field_name);
           if ($is_field_exiting) {
```
voidtek: Why do not you use the bovigo/assert function? 
In this case, the code could be:
`assert($is_field_exiting, isNotNull(), t('The field "@name" already exists and cannot be used for test purpose.', array('@name' => format_username($setting_name))));`

tests/src/Context/ContentTypeContext.php
```diff
@@ -177,8 +182,19 @@ public function aFieldGroupWithTheFollowingSettingsIsAddedToTheTypeView($arg1, T
       }
     }
 
+    $identifier = $group_machine_name . '|node|' . $arg1 . '|' . $mode;
+
+    if (field_group_exists($group_machine_name, 'node', $arg1, $mode)) {
```
voidtek: Why do not you use the bovigo/assert function?

tests/src/Context/FieldContext.php
```diff
@@ -141,7 +141,7 @@ private function scanFieldGroupsForResetting($field_groups_def_level, $default_d
 
         default:
           if (!$current_default_def_level) {
-            field_group_group_export_delete($item, FALSE);
```
voidtek: According with the documentation (http://api.drupalhelp.net/api/field_group/field_group.module/function/field_group_group_export_delete/7.2), the default value of the second parameter "$ctools_crud" is TRUE. so it is not necessary to put it here.

tests/src/Context/DrupalContext.php
voidtek: I talked to remove this tag on the page_layout feature, why do you prefer this way?

tests/features/field.multilingual.feature
voidtek: It's true. :+1:

tests/src/Context/DrupalContext.php
voidtek: This change is not done.

tests/src/Context/DrupalContext.php
voidtek: This change is not done.

tests/src/Context/ContentTypeContext.php
voidtek: I don't understand your logic, why do not you use this example (already implemented in the code):

`assert($is_field_exiting, isNotNull(), t('The field "@name" already exists and cannot be used for test purpose.', array('@name' => format_username($setting_name)))); `

tests/features/content_types/content_type_administration.feature
voidtek: I don't talk @javascript, here.

tests/features/field.multilingual.feature
voidtek: This change is not done.

tests/features/field.multilingual.feature
voidtek: Ok, legacy.

tests/features/content_types/content_type_administration.feature
voidtek: Ok, legacy.

tests/features/content_types/content_type_administration.feature
voidtek: Ok.

tests/features/content_types/content_type_administration.feature
voidtek: Ok, legacy.

tests/features/content_types/content_type_administration.feature
voidtek: Ok, legacy.

tests/src/Context/DrupalContext.php
voidtek: I don't understand your logic, why do not you use this example (already implemented in the code):

`assert($is_field_exiting, isNotNull(), t('The field "@name" already exists and cannot be used for test purpose.', array('@name' => format_username($setting_name))));`

tests/src/Context/ContentTypeContext.php
voidtek: Ok.

tests/src/Context/FieldContext.php
voidtek: If $field_name refer a deleted field and we call the function field_info_field(). What is the result?

tests/src/Context/FieldContext.php
voidtek: Ok.

tests/src/Context/FieldContext.php
voidtek: Ok.

tests/src/Context/FieldContext.php
voidtek: Ok.

tests/src/Context/ContentTypeContext.php
gillesdeudon: I say no.

tests/src/Context/DrupalContext.php
gillesdeudon: i did noit get your comment in fact. :-)

tests/src/Context/FieldContext.php
gillesdeudon: That's the question I put on the page but in my case, I got data from there with delete = 1

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
gillesdeudon: I was a mistake but after reading this doc: Your context class is initialized before each scenario is run, and every scenario has its very own context instance. It appears that this line is useless.

So, I will keep it like that.

tests/src/Context/ContentTypeContext.php
gillesdeudon: :+1:

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,219 @@
...
+        node_type_delete($result->type);
+      }
+    }
+    $this->defaultNodeTypes = array();
+  }
+
+  /**
+   * Add a field to a specified content type.
+   *
+   * @param string $arg1
```
imanoleguskiza: We should check that the content type $arg1 exists

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,219 @@
...
+   *    The machine name of the content type to which attaching the field.
+   * @param TableNode $settings
+   *    Two columns table containing field settings.
+   *
+   * @Given a field with the following settings is added to the :arg1 type:
+   *
+   * @throws \InvalidArgumentException
+   *    If the field exist already in the Drupal system.
+   */
+  public function aFieldWithTheFollowingSettingsIsAddedToTheType($arg1, TableNode $settings) {
```
imanoleguskiza: If this methods aim to create fields and field_groups, I believe they should be added into the FieldContext instead.

gillesdeudon: at first time, I thought like you but in fact, we add a field and a field group to a **content type**. On this, I follow the same logic as in TaxonomyContext.

One improvement could be to implement a generic method defined in FieldContext and used in TaxonomyContext and ContentTypeContext but I did not have a clear view on all implementation possibilities depending on the entity type. **Could we think about that in a next ticket?** (follow up ticket in this case as there are not so much different implementation compared to field groups)
and I went already far outside the ticket scope.

tests/src/Context/ContentTypeContext.php
gillesdeudon: 👍

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
gillesdeudon: Agree, @dxvargas

tests/features/content_types/content_type_mgmt.feature
```diff
@@ -0,0 +1,49 @@
+@api
```
gillesdeudon: No no, "mgmt" is a common shortcut for "management". I created this file for all scenario related to the content  type life cycle and management.
I except to see more scenarios with other content types; maybe a scenario implying 2 content types...

tests/src/Context/ContentTypeContext.php
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  /**
+   * Removes any node types created after the last list node type remembered.
+   *
+   * @AfterScenario
+   */
+  public function resetNodeTypes() {
+    if (empty($this->defaultNodeTypes)) {
```
gillesdeudon: Ok.

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+    $this->defaultFields = field_info_field_map();
+  }
+
+  /**
+   * Removes any fields created by a scenario.
+   *
+   * @AfterScenario
+   */
+  public function resetFields() {
+    if (empty($this->defaultFields)) {
```
gillesdeudon: ok.

tests/src/Context/FieldContext.php
```diff
@@ -0,0 +1,151 @@
...
+    $this->defaultFieldGroups = field_group_read_groups();
+  }
+
+  /**
+   * Removes any fields group created by a scenario.
+   *
+   * @AfterScenario
+   */
+  public function resetFieldGroups() {
+    if (empty($this->defaultFieldGroups)) {
```
gillesdeudon: Ok.

dxvargas:
👍

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
voidtek: I was referring only to that single line "$this->contentTypeWasSetUp = FALSE;"
The name of this variable is not good, yet, but it's another issue.

imanoleguskiza:
👍

imanoleguskiza:
👍

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
dxvargas: Ah ok now I understand @voidtek.
Indeed, if we run the method createMediaGalleryContentType one time $testPermissionsSetUp will be true in all subsequent AfterScenarios and method resetMediaGalleryPermission will not return in first test.

@gillesdeudon can you put back "$this->contentTypeWasSetUp = FALSE;"? It is important to be there.

profiles/common/modules/features/ec_embedded_image/ec_embedded_image.behat.inc
```diff
@@ -86,34 +86,21 @@ class ECEmbeddedImageSubContext extends DrupalSubContextBase {
...
    * @AfterScenario
    */
-  public function removeMediaGalleryContentType() {
-    if (!$this->contentTypeWasSetUp) {
+  public function resetMediaGalleryPermission() {
+    if (!$this->testPermissionsSetUp) {
       return;
     }
 
-    $this->contentTypeWasSetUp = FALSE;
```
dxvargas: Now I understood what @voidtek ment.

If we run the method createMediaGalleryContentType one time $testPermissionsSetUp will be true in all subsequent AfterScenarios and method resetMediaGalleryPermission will not return in first test.

@gillesdeudon can you put back "$this->contentTypeWasSetUp = FALSE;"? It is important to be there.

gillesdeudon: @dxvargas , 
I have a question.
When we run all scenarios, do we use the same class instance through all scenarios or a class instance is created for each scenario?
According to what I understood from web articles about the question, the answer is "an instance by scenario" and then, at each scenario, the value will be false (default value of the class attribute).

If I implement what you ask, Then, that means that according to you, we use the same instance through all scenarios? Could you please confirm that before I start to touch my code.

gillesdeudon: Line added as required... even, nobody answered to my question :-( :-(

dxvargas: I made some tests.
The class instance persists during the same Feature in different Scenarios.
The class instance does not persist for different Features.

So, I think we have a problem here. The roles that are currently being deleted after first scenario should remain until the last scenario in the feature is run.

Anyway, conclusions are:
# Class persist between scenarios in the same feature. I think `$this->contentTypeWasSetUp = FALSE;` should stay because it is out of scope to remove it and because by removing it we delete permissions twice.
# We should create a ticket to fix this problem of removing permissions right after first scenario.

dxvargas:
👍

imanoleguskiza:
👍

PR #1147, opened 2017-01-10
NEPT-173: Do not enable apachesolr_multisitesearch by default

gervasek:
MULTISITE-13875 - NEPT-173


Fefaine:
:+1:

Fefaine:
@jfhovinne I think we can give a + 1 to this,  however I would first have NEPT-519 in 2.2

PR #1158, opened 2017-01-12
NEPT-441: Change external markup validation service with the Tidy PHP extension

dxvargas:
test(interoperability): change external markup validation service with the Tidy PHP extension - NEPT-441

**We will have [Tidy](http://php.net/manual/en/book.tidy.php) as a PHP dependency and an error will happen if this extension is not installed**. This information must be visible in the commit message and release notes.


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
```diff
@@ -227,35 +227,23 @@ class TMGMTPoetryMockSubContext extends RawDrupalContext implements DrupalSubCon
...
-    $errors = $validation->messages;
+    $document = base64_decode($this->translationRequest['data']['content']);
 
-    if (!empty($errors)) {
-      $first_error = reset($errors);
+    $tidy = new tidy();
+    $tidy->parseString($document);
+    if (!$tidy->diagnose()) {
+      $errors = $tidy->errorBuffer;
       throw new Exception(
```
voidtek: Could use the assert function.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
```diff
@@ -227,35 +227,23 @@ class TMGMTPoetryMockSubContext extends RawDrupalContext implements DrupalSubCon
    * @Then the translation request document is valid XHTML
    */
   public function assertTranslationRequestDocumentIsValidXhtml() {
-    $this->assertTranslationRequest();
 
-    $document = base64_decode($this->translationRequest['data']['content']);
+    if (!class_exists('tidy')) {
+      throw new Exception('You need Tidy extension to test if XHTML is valid.');
```
voidtek: Could use the assert function.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
```diff
@@ -228,25 +228,19 @@ class TMGMTPoetryMockSubContext extends RawDrupalContext implements DrupalSubCon
    */
   public function assertTranslationRequestDocumentIsValidXhtml() {
 
-    if (!class_exists('tidy')) {
-      throw new Exception('You need Tidy extension to test if XHTML is valid.');
-    }
+    $msg = 'You need Tidy extension to test if XHTML is valid.';
+    assert(class_exists('tidy'), equals(TRUE), $msg);
```
voidtek: Why don't use the isTrue() predicate?
https://github.com/mikey179/bovigo-assert#istrue

voidtek:
:+1:

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
dxvargas: To avoid yet another use function entry in the top.

drupol: What is the problem of having an extra "use function" entry at the top ?
The overhead added by an extra "use function" is not that huge.
I really think it's more important to have easier tests to understand than avoiding the use of an extra "use function" line.

drupol:
👎

drupol:
I just added a notice regarding avoiding the use of an extra "use function" line.

Once this is fixed, I'll "thumbup" this issue.

composer.json
```diff
@@ -6,6 +6,7 @@
     "ext-json": "*",
     "ext-phar": "*",
     "ext-xml": "*",
+    "ext-tidy": "^2.0",
```
drupol: Where did you get the version number from ?

drupol:
I also think that we should keep the use of the external service to validate the markup as a fallback if the Tidy extension is not installed by default.

voidtek:
@drupol I think that the idea here, it's to remove the external service, for two reasons:
1- This is the source of random errors during CI.
2- This is an improvement for the platform to be able to run the tests without the network.

composer.json
dxvargas: By running `$ composer require ext-tidy`.

dxvargas:
@drupol I also thought about that.
But to have the same behavior in all instances is the most important requirement. The outcome should not depend on the system you're using, if it has different validation tools.

drupol:
Ok fine then.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry_mock/tmgmt_poetry_mock.behat.inc
dxvargas: For me `equals(TRUE)` is really easy to understand.
But I will change that and let's move forward.

drupol: Ok, let's use the proper function then. In the meantime, I'll thumb this up.

drupol:
:+1:

drupol:
:+1:

voidtek:
:+1:

PR #1177, opened 2017-01-24
NEPT-118 - Resolve the conflict of the TMGMT translators.

netlooker:
Main goal was to remove the tmgmt_poetry module functionalities from the tmgmt_workbench module and make it possible to use other translators which are available for the tmgmt module.


netlooker:
I'm closing this PR because performed tasks are foreseen for the release 2.4 and to make it more readable.

PR #1179, opened 2017-01-24
NEPT-118: Resolve the conflict of the TMGMT translators.

netlooker:
Main goal was to remove the tmgmt_poetry module functionalities from the tmgmt_workbench module and make it possible to use other translators which are available for the tmgmt module.

### DIFF
--- a/profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
+++ b/profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
@@ -116,12 +116,6 @@ function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form
   if ($entity_type == 'node') {
     $node = $form_state['entity'];
     if (workbench_moderation_node_moderated($node)) {
-      // If the tmgmt_poetry module is disabled remove possibility to select
-      // multiple languages and by doing that follow the default tmgmt path.
-      if (!module_exists('tmgmt_poetry')) {
-        $form['languages']['#multiple'] = FALSE;
-      }
-
       $states = array_values(_tmgmt_workbench_available_states());
       $state = $node->workbench_moderation['current']->state;
       $translation_possible = in_array($state, $states, 'strict');
--- a/profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
+++ b/profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
@@ -110,6 +110,8 @@ function tmgmt_workbench_query_tmgmt_entity_get_translatable_entities_alter(Quer
 
 /**
  * Implements hook_form_FORM_ID_alter().
+ *
+ * Alter function for the form on the 'Translate' tab.
  */
 function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form_state) {
   $entity_type = $form_state['entity_type'];
@@ -151,7 +153,11 @@ function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form
             '#disabled' => TRUE,
           );
           $form['languages']['#suffix'] = '<div class="description">' .
-            t('This content cannot be translated.') . '</div>';
+            t('Current moderation state does not allow to request a translation
+             for this content.') . '</div>';
+
+          // Remove submit button from the form.
+          unset($form['actions']);
 
           if ($always_show_direct_translation_links && ($lang_code != $node->translations->original)) {
             $options[$lang_code][TMGMT_WORKBENCH_REQUEST_OPERATION_INDEX] = '';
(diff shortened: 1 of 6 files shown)


profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
```diff
@@ -113,66 +116,37 @@ function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form
...
 
-      /** @var TMGMTJobItem[] $existing_jobs */
-      $existing_jobs = tmgmt_job_item_load_latest(
-        'workbench_moderation',
-        $entity_type,
-        $node->nid,
-        $node->translations->original
-      );
+      // Getting information about existing job items.
+      if (!isset($form_state['tmgmt_workbench']['existing_jobs'])) {
```
dxvargas: I was expecting this form_state to be always clean.
But here you are counting that a value for ['tmgmt_workbench']['existing_jobs'] may exist already.

When would ['tmgmt_workbench']['existing_jobs'] have values? Just to make sure that we really want to use that, instead of using "tmgmt_job_item_load_latest".

I also see that we have two similar functions:
tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter
tmgmt_poetry_form_tmgmt_entity_ui_translate_form_alter

In the later we have the same code here... so I guess this is correct but please can you clarify?

By the way, to imagine that these two hooks will run, no specific order... OMG!

profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
```diff
@@ -113,66 +116,37 @@ function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form
   if ($entity_type == 'node') {
     $node = $form_state['entity'];
     if (workbench_moderation_node_moderated($node)) {
+      // If the tmgmt_poetry module is disabled remove possibility to select
+      // multiple languages and by doing that follow the default tmgmt path.
```
dxvargas: Here if tmgmt_poetry is not active we are actually changing the default tmgmt behavior, which is to set $form['languages']['#multiple'] = TRUE;

So the comment is misleading.

profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
dxvargas: Ok so after talking @netlooker I understood that this is because the same information can be run twice in both mentioned hooks, but we only need this to be run once. I suggest to comment, at least for me it would have helped...

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -711,70 +709,420 @@ function tmgmt_poetry_form_tmgmt_entity_ui_translate_form_alter(&$form, &$form_s
...
 
-        // Force form to allow user actions.
-        $form_state['force_actions'] = TRUE;
+        // If translatable languages are available add the button and other
+        // elements for the additional languages requests functionality.
+        if (_tmgmt_poetry_check_translatability($form_state['tmgmt_cart']['item_id'])
+          && !empty(_tmgmt_poetry_get_translatable_languages($existing_jobs, $form, $form_state))
+        ) {
+          // Getting additional translatable languages.
+          $trans_langs = _tmgmt_poetry_get_translatable_languages($existing_jobs, $form, $form_state);
```
imanoleguskiza: I think we should define the trans_langs  variable inside the if statement directly to avoid calling the same function twice.

profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
imanoleguskiza: This is indeed weird, I don't think we need to check if tmgmt_poetry exists or not. If it is enabled the form will have the #multiple set to TRUE or FALSE if its not.

profiles/common/modules/custom/tmgmt_workbench/tmgmt_workbench.module
```diff
@@ -113,66 +116,37 @@ function tmgmt_workbench_form_tmgmt_entity_ui_translate_form_alter(&$form, $form
...
-      // NEXTEUROPA-12429 - additional header with last change date.
-      $form['languages']['#header'][TMGMT_WORKBENCH_REQUEST_LAST_CHANGE_INDEX] = t('Last change');
-      foreach ($form['languages']['#options'] as $key => $src_lang) {
-        $form['languages']['#options'][$key][] = '';
-        break;
-      }
 
       foreach ($lang_codes as $lang_code) {
-        // NEXTEUROPA-12429 - additional column with last change date.
-        // By default the last change date column is empty.
```
imanoleguskiza: HI  @netlooker, is there any particular reason we remove the references to the tickets? By no means a blocker, just curious about it.

netlooker: @imanoleguskiza I removed them to increase readability of the code before refactoring. Initial code has been moved (not only from one file to another but also across the functions) and I thought that leaving comments could be a bit misleading.

netlooker:
After investigation regarding the tmgmt module functionality I've noticed that we have an overlapping part in our tmgmt_workbench. Performed tests revealed that the custom submit and validation functions implemented for the 'Translate' tab form were not needed.
In the following commit I removed them https://github.com/ec-europa/platform-dev/pull/1179/commits/a5f30c9cbac88e23331f315a7dcffb7c602d577f

imanoleguskiza:
👍

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -711,70 +709,418 @@ function tmgmt_poetry_form_tmgmt_entity_ui_translate_form_alter(&$form, &$form_s
...
-
-        if ($translation_possible) {
-          // Add the new submit button.
+      }
+      else {
+        // Provide 'Request translation' action or hide actions.
+        // NEXTEUROPA-7719 - Simplifying Poetry UI - as a part of this ticket
+        // there is new #markup element which will be displayed every time when
+        // there is a translation request on going.
+        if ($translation_possible && !$existing_jobs) {
```
dxvargas: $existing_jobs is empty here, no need to check (see comment above).

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -711,70 +709,418 @@ function tmgmt_poetry_form_tmgmt_entity_ui_translate_form_alter(&$form, &$form_s
...
               '#disabled' => TRUE,
             ];
           }
-          $form['languages_description'] = [
-            '#type' => 'item',
-            '#prefix' => '<div class="messages warning">',
-            '#markup' => t('Please wait the acceptation translation process before update request.'),
-            '#suffix' => '</div>',
-          ];
+          if ($existing_jobs && !isset($form['languages_description'])) {
```
dxvargas: $existing_jobs is empty here, so what is inside will never run (see comment above).

tests/features/tmgmt/tmgmt_workbench.feature
```diff
@@ -158,4 +158,3 @@ Feature: TMGMT Workbench features
     And I select "Translator A" from "Translator"
     And I press "Submit to translator"
     Then I should not see the text "This content cannot be translated."
-    And I should see the text "Please end up the active translation process before creating a new request."
```
netlooker: We need to check if in the tmgmt_poetry this case is covered. Main reason for removing that step was to not to have a dependency from the tmgmt_poetry inside tmgmt_workbench.

dxvargas:
👍

imanoleguskiza:
👍

PR #1183, opened 2017-01-26
NEPT-567: Prevent wrong missing module when rebuilding registry - fix(maintenance)

Fefaine:
…y NEPT-567


resources/multisite_drupal_standard.make
```diff
@@ -242,6 +242,10 @@ projects[feature_set][patch][] = patches/feature_set-misc-nexteuropa_4459.patch
 
 projects[feeds][subdir] = "contrib"
 projects[feeds][version] = "2.0-beta1"
+; Issue #2828605: Feature set does not invoke hook_requirements().
```
netlooker: I think we have a small description error here.

gillesdeudon:
:+1:

imanoleguskiza:
👍

PR #1200, opened 2017-02-01
NEPT-611: Fix display of username in watchdog upon logout. - support-2.2 - fix(maintenance)

drupol:


drupol:
Test on Jenkins is green.

Fefaine:
:+1:

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -213,7 +213,7 @@ function ecas_login_page() {
 function ecas_logout($full_logout = FALSE) {
   global $user;
 
-  watchdog('user', 'Session closed for %name.', array('%name' => theme('placeholder', $user->name)));
+  watchdog('user', 'Session closed for %name.', array('%name' => format_username($user)));
```
dxvargas: If using format_username, as documentation advises, check_plain should be used wrapping the returned string.

But I would just use $user->name, because it is more straight forward and also because it is the way it is used in function user_logout_current_user and for me to use same logic for same results is a plus.

Anyway, with or without format_username, we need a change in code here.

drupol: After some check and confirmation by @Fefaine, there is no need to have a `check_plain()` here, the watchdog will do it for us.

dxvargas:
👍

drupol:
Dear @dxvargas and @Fefaine, using `%name` as placeholder is a valid solution, see: https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/format_string/7.x followed by https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_placeholder/7.x.

`%name` strings are passed through [drupal_placeholder()](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_placeholder/7.x) which calls [check_plain()](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/check_plain/7.x).

dxvargas:
👍

netlooker:
:+1:

drupol:
Fix tag has been added too.

PR #1203, opened 2017-02-02
NEPT-548: Prevents the change of e-mail addresses of connected users on contact forms.

drupol:
Prevents the change of e-mail addresses of connected users when they are on the contact forms

### DIFF
--- a/tests/features/contact_form.feature
+++ b/tests/features/contact_form.feature
@@ -35,7 +35,7 @@ Feature: Contact Form
       | Your message has been sent. |
 
   @javascript
-  Scenario: Administrator user can submit the contact page
+  Scenario: Logged in users are not able to edit their name and email.
     Given I am logged in as a user with the administrator role
     And the module is enabled
       | modules      |
--- a/tests/features/contact_form.feature
+++ b/tests/features/contact_form.feature
@@ -12,8 +12,6 @@ Feature: Contact Form
       | contact_form |
     When I am on "contact"
     Then I should see "Contact - European Commission" in the "title" tag
-    When I fill in "Your name" with "Chuck"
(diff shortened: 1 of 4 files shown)


tests/features/contact_form.feature
```diff
@@ -12,8 +12,6 @@ Feature: Contact Form
       | contact_form |
     When I am on "contact"
     Then I should see "Contact - European Commission" in the "title" tag
-    When I fill in "Your name" with "Chuck"
```
gillesdeudon: The test is not correct anymore because an anonymous user **can** insert a name and an e-mail in the contact form. 
So, we must reinsert these 2 lines

tests/features/contact_form.feature
```diff
@@ -29,8 +27,6 @@ Feature: Contact Form
...
-    And I fill in "Your e-mail address" with "chuck.norris@improbabledommainname.com"
```
gillesdeudon: The tests are incomplete because now, if the user is logged, the test must ensure that any authenticated user cannot type a name and an e-mail in the form and that he views its data instead.

tests/features/contact_form.feature
imanoleguskiza: I agree with @gillesdeudon this two lines need to be present because we need to make sure that anonymous users can add their name and email

drupol: Ok it's done.

tests/features/contact_form.feature
drupol: We agree to discard this comment.

gillesdeudon:
:+1:

dxvargas:
👍

PR #1216, opened 2017-02-07
NEPT-316: Implement the solution for testing the mail features

netlooker:
Implemented solution is provided via new custom Behat DrupalMailContext context.
It doesn't need any additional modules or settings to run it.

### DIFF
--- a/tests/features/drupal_mail.feature
+++ b/tests/features/drupal_mail.feature
@@ -1,0 +1,20 @@
+@api @internalMail
+Feature: Drupal Mail features
+  In order to check if a mail was sent correctly
+  As an Administrator
+  I want to be able to catch the mail and verify its properties
+
+  Background:
+    Given platform is configured to use the internal mail handling
+
+  Scenario: Checking mail action for requesting new password.
+    When I go to "/user"
+    And I click "Request new password" in the "primary_tabs" region
+    And I fill in "Username or e-mail address" with "admin"
+    And I press "E-mail new password"
+    Then I should see the success message "Further instructions have been sent to your e-mail address."
+    And the internal mail handler received the mail
+    And the captured mail has the following properties:
+      | from        | admin@example.com                                     |
+      | to          | admin@example.com                                     |
+      | subject     | Replacement login information for admin at NextEuropa |
--- a/tests/features/drupal_mail.feature
+++ b/tests/features/drupal_mail.feature
@@ -6,15 +6,18 @@ Feature: Drupal Mail features
 
   Background:
     Given platform is configured to use the internal mail handling
+    And users:
+      | name  | mail         | roles        |
+      | foo   | foo@bar.com  | contributor  |
 
-  Scenario: Checking mail action for requesting new password.
+  Scenario: Checking mail action for requesting new password by using the username as identifier.
     When I go to "/user"
     And I click "Request new password" in the "primary_tabs" region
-    And I fill in "Username or e-mail address" with "admin"
+    And I fill in "Username or e-mail address" with "foo"
     And I press "E-mail new password"
     Then I should see the success message "Further instructions have been sent to your e-mail address."
-    And the internal mail handler received the mail
-    And the captured mail has the following properties:
+    And the e-mail has been sent
+    And the sent e-mail has the following properties:
       | from        | admin@example.com                                     |
-      | to          | admin@example.com                                     |
-      | subject     | Replacement login information for admin at NextEuropa |
+      | to          | foo@bar.com                                           |
+      | subject     | Replacement login information for foo at NextEuropa |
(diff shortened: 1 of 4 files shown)


tests/src/Context/DrupalMailContext.php
```diff
@@ -0,0 +1,163 @@
...
+    $this->variableContext->setVariable('mail_system', $mail_system);
+  }
+
+  /**
+   * Verifies that Drupal internal mail handler received the mail.
+   *
+   * @Then the internal mail handler received the mail
+   */
+  public function assertInternalMailHandlerReceivedTheMail() {
+    $captured_mail = self::getCapturedMail();
```
gillesdeudon: I would use `$this->` instead of `self::` because `getCaptureMail()` manages non-static parameter; itself, it is not static.

tests/features/drupal_mail.feature
```diff
@@ -0,0 +1,20 @@
+@api @internalMail
+Feature: Drupal Mail features
+  In order to check if a mail was sent correctly
+  As an Administrator
+  I want to be able to catch the mail and verify its properties
+
+  Background:
+    Given platform is configured to use the internal mail handling
+
+  Scenario: Checking mail action for requesting new password.
```
gillesdeudon: I would change  "_Checking mail action for requesting new password._" for "_Checking mail action for requesting new password **by using the username as identifyer**._"

tests/features/drupal_mail.feature
```diff
@@ -0,0 +1,20 @@
...
+    When I go to "/user"
+    And I click "Request new password" in the "primary_tabs" region
+    And I fill in "Username or e-mail address" with "admin"
+    And I press "E-mail new password"
+    Then I should see the success message "Further instructions have been sent to your e-mail address."
+    And the internal mail handler received the mail
+    And the captured mail has the following properties:
```
gillesdeudon: I would replace the sentence "_And the internal mail handler received the mail_" by "_And the e-mail is sent_" which is more understandable from the Business perspective.

Same for "_And the captured mail has the following properties_", I would write "_And the sent e-mail has the following properties_".

tests/features/drupal_mail.feature
```diff
@@ -0,0 +1,20 @@
...
+      | from        | admin@example.com                                     |
+      | to          | admin@example.com                                     |
```
gillesdeudon: I would create a specific user for the _to_ in order to ensure that the feature does not use site's one by mistake what could happen in some buggy cases.

netlooker:
@gillesdeudon thanks for the feedback. I've implemented your suggestions.

netlooker:
@gillesdeudon I've changed the scenario scope description according to the talk we had.

gillesdeudon:
:+1:

Fefaine:
@netlooker we decided to not add a log for 2.4 can you remove the entry?

tests/src/Context/DrupalMailContext.php
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Verifies that Drupal internal mail handler received the mail.
+   *
+   * @Then the e-mail has been sent
+   */
+  public function assertInternalMailHandlerReceivedTheMail() {
+    $captured_mail = $this->getCapturedMail();
+
+    if (empty($captured_mail)) {
```
voidtek: Why do you not use the assert function of bovigo?

tests/src/Context/DrupalMailContext.php
```diff
@@ -0,0 +1,163 @@
...
+   * Assert the properties in the captured mail.
+   *
+   * @Then the sent e-mail has the following properties:
+   */
+  public function assertCapturedMailHasTheFollowingProperties(TableNode $table) {
+    $this->assertCapturedMail();
+    $expected_properties = $table->getRowsHash();
+
+    foreach ($expected_properties as $key => $value) {
+      assert($this->capturedMail[$key], equals($value));
```
voidtek: For a business use, should be better to add a message of this assert. What does you think?

Actually, the exception output is :

      Failed asserting that 'admin@example.com' is equal to <string:admin@examplee.com>.
      --- Expected
      +++ Actual
      @@ @@
      -'admin@examplee.com'
      +'admin@example.com'
       (bovigo\assert\AssertionFailure)

gillesdeudon:
:+1:

tests/src/Context/DrupalMailContext.php
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /**
+   * Returns a mail captured by the Drupal internal mail handler.
+   *
+   * @return array
+   *   An empty or a mail data array
+   */
+  private function getCapturedMail() {
+    // During the first run with the clean state the variable_get function
```
netlooker: It is because I'm not setting up `drupal_test_email_collector` variable like in the given example. If you think that we could do that please consider adding the small change to the function platformIsConfiguredToUseInternalMailHandling().

dxvargas:
👍

gillesdeudon:
:+1:

PR #1222, opened 2017-02-10
NEPT-421: Prevent errors when apachesolr module is not enabled.

drupol:


profiles/common/modules/custom/multisite_drupal_toolbox/multisite_drupal_toolbox.module
```diff
@@ -807,28 +807,13 @@ function _disable_block($module, $delta) {
...
+  if (!module_exists('apachesolr')) {
     return;
   }
-  else {
-    $info = new stdClass();
-    $info->type = $bundle;
-  }
 
   // Update hook cache to find new hooks (see module_implements()).
   cache_set('module_implements', array(), 'cache_bootstrap');
```
gillesdeudon: Should that be `clear_cache_all('module_implments', 'cache_bootstrap');` because:

- We just want to refresh the cache after;
- If the expire setting of this cache record is  PERMANENT, the [doc](https://api.drupal.org/api/drupal/includes!cache.inc/function/cache_set/7.x) says : 

> CACHE_PERMANENT: Indicates that the item should never be removed unless explicitly told to using cache_clear_all() with a cache ID.

drupol: I will investigate this morning.

drupol: Based on [the function signature](https://api.drupal.org/api/drupal/includes%21cache.inc/function/cache_clear_all/7.x), it makes sense indeed.

I will create a new ticket, unless you think it makes sense to fix this here.

drupol: Apparently it's done multiple time in the module and [once in Drupal core](http://cgit.drupalcode.org/drupal/tree/includes/module.inc?h=7.54#n741).

gillesdeudon:
:+1:

drupol:
For a reason that we do not know, Jenkins didn't returned the status. However, the tests are green.

dxvargas:
👍

PR #1224, opened 2017-02-10
NEPT-132: Update contrib module Metatag.

drupol:


drupol:
See #1231 for support-2.2 branch pull request.

voidtek:
:+1:

profiles/common/modules/features/nexteuropa_metatags/nexteuropa_metatags.helpers.inc
```diff
@@ -0,0 +1,27 @@
...
+ *   Returns the error message if the config of metatags does not comply.
+ */
+function _nexteuropa_metatags_check_metatag_config() {
+  $messages = array();
+
+  if ($config = metatag_config_load('global')) {
+    $meta_to_check = array('X-UA-Compatible', 'viewport');
+
+    foreach ($meta_to_check as $meta) {
+      if (!isset($config->config[$meta]) || (isset($config->config[$meta]) && empty($config->config[$meta]['value']))) {
```
gillesdeudon: What if the sub-sites has overridden the ec_resps theme and remove the metag from their html.tpl.html? In this condition, they need to set the value and they will receive this message all the time, even it is not pertinent.

I see 2 solutions:

- Complex: Scanning the html.tpl.php to see if the meta tag is there;
- Simple: Modifying the message to say more or less: "if you have a specific "html.tpl.php" that do not contain this meta tag, forget this warning"

Fefaine: If the user has remove the meta X-UA-Compatible from his tpl there is no issue, the meta is now provided by the module. 
If the user has that meta tag in his template, he sould now remove it.
If the user has no meta tag in this template and config then he gets the error. Therefore for me it's perfect like this

Fefaine:
:+1:

PR #1231, opened 2017-02-16
NEPT-132: Metatag contrib security update.

drupol:
Also add a warning if the X-UA-Compatible meta is set by the user.

### DIFF
--- a/profiles/common/modules/features/nexteuropa_metatags/nexteuropa_metatags.helpers.inc
+++ b/profiles/common/modules/features/nexteuropa_metatags/nexteuropa_metatags.helpers.inc
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @file
+ * Code for the NextEuropa Metatags module.
+ */
+
+/**
+ * Custom helper function.
+ *
+ * @return array
+ *   Returns the error message if the config of metatags does not comply.
+ */
+function _nexteuropa_metatags_check_metatag_config() {
+  $messages = array();
+
+  if ($config = metatag_config_load('global')) {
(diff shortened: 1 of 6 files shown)


voidtek:
:+1:

profiles/common/modules/features/nexteuropa_metatags/nexteuropa_metatags.helpers.inc
```diff
@@ -0,0 +1,27 @@
...
+ */
+function _nexteuropa_metatags_check_metatag_config() {
+  $messages = array();
+
+  if ($config = metatag_config_load('global')) {
+    $meta_to_check = array('X-UA-Compatible');
+
+    foreach ($meta_to_check as $meta) {
+      if (isset($config->config[$meta]) && !empty($config->config[$meta]['value'])) {
+        $messages[$meta] = t('The <em>@meta</em> global meta is not empty, please <a href="@link">configure it</a> and set it to empty or else you might end up in duplicates meta tags in your HTML.', array('@meta' => $meta, '@link' => url('admin/config/search/metatags/config/global')));
```
gillesdeudon: What if the sub-sites has overridden the ec_resps theme and remove the metag from their html.tpl.html? In this condition, they need to set the value and they will receive this message all the time, even it is not pertinent.

I see 2 solutions:

- Complex: Scanning the html.tpl.php to see if the meta tag is there;
- Simple: Modifying the message to say more or less: "if you have a specific "html.tpl.php" that do not contain this meta tag, forget this warning"

Fefaine:
:+1:

PR #1258, opened 2017-03-06
Analysis: Cannot install site with the lock of cce_basic_config feature

gillesdeudon:


gillesdeudon:
Obsolote PR I close it

PR #1282, opened 2017-03-10
NEPT-775: Invalidate the cache by default every time a page is created / updated 

netlooker:


profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.admin.inc
```diff
@@ -24,6 +24,67 @@ function nexteuropa_varnish_admin_settings_form($form, $form_state) {
...
+      $state = t('off');
+      break;
+
+    case 1:
+      $state = t('on');
+      break;
+  }
+
+  if ($rule_state != $rule_state_var) {
+    variable_set('nexteuropa_varnish_default_purge_rule', $rule_state);
```
gillesdeudon: It is already done by the default admin submit, why do we have to set the variable again?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.admin.inc
```diff
@@ -24,6 +24,67 @@ function nexteuropa_varnish_admin_settings_form($form, $form_state) {
     '#default_value' => variable_get('nexteuropa_varnish_default_purge_rule', FALSE),
   );
 
+  $form = system_settings_form($form);
+  array_unshift($form['#submit'], '_nexteuropa_varnish_default_rule_state_check');
+
   return system_settings_form($form);
```
gillesdeudon: I tested, if you make 
`$form = system_settings_form($form);`
...
`return system_settings_form($form);`
, the form is submitted twice.

`$form = system_settings_form($form);`
...
`return $form;`, do the trick.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.admin.inc
```diff
@@ -24,6 +24,67 @@ function nexteuropa_varnish_admin_settings_form($form, $form_state) {
     '#default_value' => variable_get('nexteuropa_varnish_default_purge_rule', FALSE),
   );
 
+  $form = system_settings_form($form);
+  array_unshift($form['#submit'], '_nexteuropa_varnish_default_rule_state_check');
```
gillesdeudon: Why must it be executed before the actual saving of the settings? And why do we need another message; the default message "The configuration option have been saved" is not enough?
There is only one option; its checkbox and its label are obvious enough to avoid an additional step in the form submit process, no?

gillesdeudon:
:+1:

imanoleguskiza:
👍

gillesdeudon:
:+1:

PR #1287, opened 2017-03-13
NEPT-324: Add check on the varnish settings as path purge and runtime

gillesdeudon:


imanoleguskiza:
👍

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.helper.inc
```diff
@@ -0,0 +1,85 @@
...
+ * CHeck the module settings to see if they are correctly defined.
+ *
+ * @param array $settings
+ *   The module settings to check.
+ *
+ * @return bool
+ *   FALSE if at least one of the parameters is not correctly defined;
+ *   otherwise TRUE.
+ */
+function _nexteuropa_varnish_check_configuration($settings) {
```
voidtek: This function can be improved with dynamic tests:

```
$all_configurations = array('varnish_request_method', 'varnish_http_targets', 'varnish_tag', 'varnish_request_user', 'varnish_request_password', 'varnish_http_timeout');
foreach ($all_configurations as $configuration) {
	if (empty($settings[$configuration])) {
		return FALSE;
	}
}
return TRUE;
```

This remark is not blocker.

tests/src/Context/DatabaseLogContext.php
```diff
@@ -65,4 +84,31 @@ protected function assertMessageLogged($severity, $type, $message) {
...
+   *   The severity of the message.
+   * @param string $type
+   *   The type (category) of the message.
+   * @param string $message
+   *   A regular expression pattern the message needs to match.
+   *
+   * @see watchdog()
+   */
+  protected function assertMessageNotLogged($severity, $type, $message) {
+    $query = db_select('watchdog', 'w');
```
voidtek: We could create a function with this part of code because it's used on two functions.(assertMessageLogged and assertMessageNotLogged)

tests/src/Context/VariableContext.php
```diff
@@ -42,6 +42,22 @@ public function setVariable($name, $value) {
...
 
   /**
+   * Deletes the value of a Drupal variable.
+   *
+   * The initial value of the value is remembered for later restore.
+   *
+   * @param string $name
+   *   Name of the variable.
+   */
+  public function deleteVariable($name) {
```
voidtek: The name of this function should be unsetVariable().
deleteVariable() is the opposite of addVariable().

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -227,6 +229,25 @@ function _nexteuropa_varnish_merge_paths_from_purge_rules($purge_rules, $node) {
  * Purges certain paths on the web frontend cache.
  */
 function _nexteuropa_varnish_purge_paths($paths) {
+  try {
+    $settings = _nexteuropa_varnish_get_varnish_settings();
+  }
+  catch (InvalidArgumentException $iae) {
+    $end_sentence = t('Please ask your support team to check the server configuration.');
```
voidtek: We could create a function with this part of code because it's used on two functions (we need to add a parameter for the beginning of the message).
(_nexteuropa_varnish_purge_paths & _nexteuropa_varnish_base_request_headers)

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.install
```diff
@@ -37,27 +39,37 @@ function nexteuropa_varnish_schema() {
...
+    $settings = _nexteuropa_varnish_get_varnish_settings(FALSE);
 
     foreach ($required_vars as $var) {
-      if (!isset($conf[$var])) {
+      if (!isset($settings[$var])) {
+        $t_var = array('@var' => $var);
         $requirements[$var] = array(
-          'description' => $t('Following variable must be set: @var.', array('@var' => $var)),
+          'title' => $t('NextEuropa Varnish: "nexteuropa_@var" is missing', $t_var),
+          'description' => $t('Following variable must be set: nexteuropa_@var. Please ask your support team to check the server configuration.', $t_var),
```
voidtek: It's only that the support team (QA/Maintenance/Devops) which sees this message. The sentence "Please ask your support team to check the server configuration." is unnecessary.

tests/features/frontend_cache_purge.feature
```diff
@@ -510,3 +512,19 @@ Feature:
     And I fill in "Moderation notes" with "Immediately publishing this"
     And I press "Save"
     Then an informational message is logged with type "nexteuropa_varnish" and a message matching "Clearing paths: /more-basic-pages, /, /even-more-basic-pages"
+
+  Scenario: No purge are instructed if a part of the configuration has disappeared.
+    Given the following cache purge rules:
+      | Content Type | Paths to Purge         |
+      | page         | /more-basic-pages, /   |
```
voidtek: One rule is enough for this test, with the path "frontend-cache-purge-publish-immediately"

tests/src/Context/VariableContext.php
gillesdeudon: I do not get the idea here because I use "variable_del" where "del" means "delete". 

Then, the function name is aligned with what happens in its process.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.install
gillesdeudon: It is also displayed when you activate the related feature set. So, It is not so useless.

tests/features/frontend_cache_purge.feature
gillesdeudon: And? Is it a problem to have nevertheless 2 rules?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
gillesdeudon: I prefer to leave this question for a next evolution. So far, there is enough differences to justify to keep like this.
So, I will not follow this remark.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.helper.inc
gillesdeudon: :+1:

tests/src/Context/DatabaseLogContext.php
gillesdeudon: :+1:

tests/features/frontend_cache_purge.feature
voidtek: It's only confusing.
A simple rule with the path "/" is enought for this test.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.install
voidtek: It's true, you are alright.

voidtek:
:+1:

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.install
```diff
@@ -37,27 +39,37 @@ function nexteuropa_varnish_schema() {
...
 
+  if (in_array($phase, $control_phase)) {
     $required_vars = array(
-      'nexteuropa_varnish_http_targets',
-      'nexteuropa_varnish_request_method',
-      'nexteuropa_varnish_request_user',
-      'nexteuropa_varnish_request_password',
-      'nexteuropa_varnish_tag',
-      'nexteuropa_varnish_http_timeout',
+      'varnish_http_targets',
```
gillesdeudon: Done

netlooker:
:+1:

tests/features/frontend_cache_purge.feature
```diff
@@ -515,9 +515,9 @@ Feature:
 
   Scenario: No purge are instructed if a part of the configuration has disappeared.
     Given the following cache purge rules:
-      | Content Type | Paths to Purge         |
-      | page         | /more-basic-pages, /   |
-      | page         | /even-more-basic-pages |
+      | Content Type | Paths to Purge       |
+      | page         | /more-basic-pages, / |
+      | page         |                      |
```
imanoleguskiza: Hi @gillesdeudon,
I'm a bit confused with this last change. Other than this, I agree with all the rest

gillesdeudon: Hello,

I changed this part because of the remark of @voidtek. I realized that I needed 2 rules to highlight the fact whatever the number of rules, nothing is sent to Varnish if the settings are not complete.

IMO, the previous version was not pertinent and I think in an other ticket we should align the scenario above this one, because it implies 2 rules that impact other pages than the currently saved page.

- /even-more-basic-pages
- / (front page)
- /more-basic-pages

So, the test covered one rule type only.

By replacing the rule with the "/even-more-basic-pages" path by the new one that purge , I cover now the 2 possible rule types:

1.  Purge of a specific page when a page content is saved; it is the first rule described in the table;
2. Purge the currently saved page; it is the second rule of the table.

The test is then complete for me.

imanoleguskiza: Ok for me then, giving it my +1

imanoleguskiza:
👍

PR #1289, opened 2017-03-14
Nept 434 The selector for the poetry job request should not allow to select date in the past

Fefaine:


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -225,6 +226,20 @@ function tmgmt_poetry_form_tmgmt_job_form_alter(&$form, &$form_state, $form_id)
 }
 
 /**
+ * Custom validator to add a server validation step for dates in the past.
+ */
+function _tmgmt_poetry_validate($form, &$form_state) {
```
imanoleguskiza: I think the function name should be more descriptive about its function. Something like "_tmgmt_poetry_validate_job_requested_date"

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -225,6 +226,20 @@ function tmgmt_poetry_form_tmgmt_job_form_alter(&$form, &$form_state, $form_id)
...
+  $delai = $form_state['values']['settings']['delai'];
+  if (is_string($delai)) {
+    $date = strtotime($delai);
+    $today = time();
+    if ($date < $today) {
+      form_set_error('delai', t('The date you selected is incorrect!'));
```
imanoleguskiza: Something similar here, I think it would be better to add a more explicit message: "The expected requested delivery date can't be in the past."

tests/features/tmgmt/tmgmt_poetry_translations.feature
```diff
@@ -760,3 +760,28 @@ Feature: TMGMT Poetry features
...
+      | en       | Title date test  | Test delai conditions   |
+    When I click "Translate" in the "primary_tabs" region
+    When I check the box on the "French" row
+    And I press "Request translation"
+    And I fill in "Date" with "01/01/1970"
+    And I press "Submit to translator"
+    Then I should see the error message "The date you selected is incorrect!"
+
+    @javascript
+  Scenario: Request translation of a page with date in the past using javascript
```
imanoleguskiza: More a question than anything else: why two scenarios for the same thing?

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
Fefaine: Ok my idea was to have a generic name in case we add more validation afterwards, but i guess we'll just need to change the name if it happens :)

tests/features/tmgmt/tmgmt_poetry_translations.feature
Fefaine: There are two ways of blocking the selection, one js and one server. But you are right that the javascript test does not test the popup so I will remove it

imanoleguskiza:
👍

dxvargas:
👍

Fefaine:
:+1:

PR #1291, opened 2017-03-14
Revert "NEPT-323: Add content type column on recent translations jobs"

dxvargas:
Reverts ec-europa/platform-dev#1286 because it was squashed and it shouldn't.


gillesdeudon:
:+1:

PR #1309, opened 2017-03-21
NEPT-831: Cleaning the environment before running the Behat feature

netlooker:


tests/behat.yml.dist
```diff
@@ -28,6 +27,9 @@ default:
         - Drupal\nexteuropa\Context\DatabaseLogContext
         - Drupal\nexteuropa\Context\PathautoContext
         - Drupal\nexteuropa\Context\MultilingualContext
+        - Drupal\nexteuropa\Context\DrushContext:
+            drush_bin: "{{ drush.bin }}"
+            db_dump_location: '{{ drush.db.dump }}'
```
ademarco: I would add here an `enabled:` boolean property set to `false` on local environment and to `true` on Jenkins, so that we have control when dumping and restoring happens.

netlooker: Your point will be addressed in follow up ticket NEPT-852.

dxvargas:
👍

tests/features/piwik.feature
```diff
@@ -1,22 +1,20 @@
-@wip
+@api
 Feature: Check Piwik
   In order to check if the the type attribute is set for the Piwik element.
```
gillesdeudon: As you have a conflict to resolve on your branch, mayb you can take the opportinity to fix this small typo by removing the  useless additional "the" in  "check if the *the* type attribute"

voidtek:
:+1:

gillesdeudon:
:+1:

ademarco:
👍

PR #1316, opened 2017-03-22
Feature/nepcb 1

gervasek:


Fefaine:
The tests do not need to run for both profiles

PR #1330, opened 2017-03-30
Nept 252

imanoleguskiza:


profiles/common/modules/custom/nexteuropa_lastupdate/nexteuropa_lastupdate.module
```diff
@@ -0,0 +1,192 @@
...
+ *   The node to get the last published date from.
+ *
+ * @return array
+ *   An array containing the label and date to be rendered.
+ */
+function _nexteuropa_lastupdate_node_callback($node) {
+  if ($node->status) {
+    return array(
+      'label' => t('Last published'),
+      'date' => date('d/m/Y', $node->changed),
```
gillesdeudon: One QA rule is to use as much as possible Drupal functions. In this case to format it, we have [format_date](https://api.drupal.org/api/drupal/includes!common.inc/function/format_date/7.x) that should `replace date()`.

Furthermore, we should not enforce the date format like it is done. We should use those supplied by Drupal en existing in Drupal: 'long', short or medium.
Up to you to determine which one to choose.

profiles/common/modules/custom/nexteuropa_lastupdate/nexteuropa_lastupdate.module
```diff
@@ -0,0 +1,192 @@
...
+ * @param object $user
+ *   The user to get the last accessed date from.
+ *
+ * @return array
+ *   An array containing the label and date to be rendered.
+ */
+function _nexteuropa_lastupdate_user_callback($user) {
+  return array(
+    'label' => t('Last accessed'),
+    'date' => date('d/m/Y', $user->access),
```
gillesdeudon: Same comment about dates as above

Question: Don't you think that `_nexteuropa_lastupdate_comment_callback`, `_nexteuropa_lastupdate_user_callback` and `_nexteuropa_lastupdate_node_callback` are quite similar

Could we not have a central one like:
```
_nexteuropa_lastupdate_entity_last_date_info($entity) {
  $last_date = _nexteuropa_lastupdate_get_last_date($entity);
  return array(
     'label' => t('Last uploaded'),
     'date' =>$last_date,
  );
}
```

profiles/common/modules/custom/nexteuropa_lastupdate/nexteuropa_lastupdate.module
```diff
@@ -0,0 +1,192 @@
...
+ * @param object $comment
+ *   The comment to get the last published date from.
+ *
+ * @return array
+ *   An array containing the label and date to be rendered.
+ */
+function _nexteuropa_lastupdate_comment_callback($comment) {
+  return array(
+    'label' => t('Last published'),
+    'date' => date('d/m/Y', $comment->changed),
```
gillesdeudon: Same comment about dates as above.

Actually, there is no page for comments. They are integrated in a content page. So is it pertinent to foresee it?

gillesdeudon:
:+1:

gillesdeudon:
:+1:

voidtek:
:+1:

PR #1343, opened 2017-04-04
NEPT-250: Use a real example for PHPUnit

dxvargas:


ademarco:
Closing as this is taken care in the original branch now.

PR #1358, opened 2017-04-06
NEPT-860: Resolution of the conflicts for the merge of master into 2.4

gillesdeudon:


imanoleguskiza:
👍

tests/features/frontend_cache_purge.feature
```diff
@@ -553,7 +553,6 @@ Feature:
     And I press "Save"
     Then the web front end cache was not instructed to purge any paths
     And a critical error message is logged with type "nexteuropa_varnish" and a message matching "No path has been sent for clearing because all module settings are not set."
-    And no informational message is logged with type "nexteuropa_varnish" and a message matching "Clearing paths: /more-basic-pages, /, /even-more-basic-pages"
```
voidtek: I do not understand this conflict fix. I think that we introduce a regression bug with this PR.
We need to resolve the situation, and for me, it's not good to have these two messages in the same time in the logs.

gillesdeudon: I thought like you when I saw the test but after seeing the content of the message saying "Clear**ing** paths" instead of "Clear**ed** paths". Both messages are not incompatible because one informs about an intention: "I inform that I will try to send this path" while the other informs about an error: "I inform that I did not manage to send them".
Added to what I saw in the different behat tests and to the code/process reorganization I add to do to merge the master based code with the 2.4 code, that's why I decided that it was the best approach.

voidtek:
:+1:

PR #1385, opened 2017-04-21
NEPT-799: NEXTEUROPA VARNISH: the old alias of an updated content is never cleared by the process

gillesdeudon:


gillesdeudon:
I close for avoiding problem with 2.4. I will recreate a new branch and PR based on 2.4

PR #1398, opened 2017-04-28
NEPT-970: Refactoring the Europa Component views row style plugin.

drupol:
There's one issue with this plugin and I can't afford to spend more time in it, I almost sure this comes from the Drupal's form handling and I'm not able to find a working solution.

The problem is in Views. When you select the row plugin "Europa Component".
When you edit the settings of the plugin, you have to select which hook theme to use.
Then, once it's selected, the form reloads in AJAX and shows you the mapping between the variables of the theme hook and the fields available in the view.

Once you do your mapping and hit APPLY, for a reason that I do not know, the popup form will lose its settings.
Here, the mapping will be lost and you will have to redo it once again.

Another point of discussion in this module is, should we include ALL the hook theme in the SELECT options or only some specifics ones ?

By default, the theme hooks are gathered when we flush the caches. I know it's ugly, but as I said in the comment:

> AFAIK, there is no other way to do this horrendous hack to retrieve ALL the themes hooks available.
> We could use the regular way to do it with module_invoke_all('theme') and it works properly. The only issue is that theme hooks defined in the theme are not detected because we are using an admin theme. Any alternative way is very welcome.

Anyway, I'm waiting for feedback regarding these points then I'll continue my work on it.

About the name of the branch, this is a mistake, do not mind it please.


profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
```diff
@@ -0,0 +1,124 @@
...
+   * @inheritDoc
+   */
+  function render($row) {
+    static $row_index;
+    if (!isset($row_index)) {
+      $row_index = 0;
+    }
+
+    // In order to have automatic theme hooks suggestions.
+    $theme = sprintf('%s__%s__%s', $this->options['theme'], $this->view->name, $this->view->current_display);
```
netlooker: @drupol I think that this part instead of `sprintf()` function could be implemented by the `format_string()`. What do you think about it?

profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
```diff
@@ -0,0 +1,124 @@
...
+
+    // We build the variables array based on fields mapping.
+    $variables = array();
+    foreach ($this->options['variables'] as $key => $variable) {
```
netlooker: @drupol what about moving this part into the `pre_render` method? It's not something super important but we should keep rendering as clean as it's possible.

profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
```diff
@@ -0,0 +1,124 @@
...
+   */
+  public function options_form(&$form, &$form_state) {
+    parent::options_form($form, $form_state);
+
+    $hooks_theme = variable_get('nexteuropa_core_views_hooks', array());
+    $form['theme'] = array(
+      '#type' => 'select',
+      '#required' => TRUE,
+      '#title' => t('Theme hook'),
+      '#description' => t('TODO'),
```
netlooker: A short description would be appreciated.

profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
drupol: I will comment in the code.

profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
drupol: You're right, I will commit a fix.

profiles/common/modules/custom/nexteuropa_core_views/plugins/component/views_europa_component_plugin_rows.inc
netlooker: Ok. I'm still filling here not 100% but I've tried to attack this and it seems that this implementation is strong enough :)

drupol:
Closed this in favor of: #1400

PR #1399, opened 2017-04-28
NEPT-631: Added first version of validation

imanoleguskiza:


profiles/common/modules/features/cce_basic_config/cce_basic_config.module
```diff
@@ -341,3 +360,54 @@ function cce_basic_config_modules_enabled($modules) {
...
+    // Get mime type from filename.
+    $mime_by_filename = file_get_mimetype($file->filename);
+    // Get mime type from file contents.
+    $finfo = finfo_open(FILEINFO_MIME_TYPE);
+    // Drupal will in some specific cases append a .txt extension to a file
+    // which will mess with the temporary PHP uri.
+    $temporary_uri = explode(".", $file->uri);
+    $real_mime = finfo_file($finfo, $temporary_uri[0]);
+    if ($mime_by_filename !== $real_mime) {
+      $errors[] = "The file's content type doesn't match the extension.";
```
gillesdeudon: I suggest as message: 

> The file's extension ("[extension]") does not match the file mime type ("[mimetype]")
where:

- [extension] is the extension retrieved from `$file->filename`;
- [mimetype] is the $real_mime value.

We should be as verbose as possible to inform the user about the error.

I think also that we should log the error (watchdog).

profiles/common/modules/features/cce_basic_config/cce_basic_config.module
```diff
@@ -341,3 +360,54 @@ function cce_basic_config_modules_enabled($modules) {
...
+}
+
+/**
+ * Helper function to check if a user has access to the destination file system.
+ */
+function _cce_basic_config_validate_file_destination(&$errors, $file) {
+  if (variable_get("ec_enable_file_destination_validation", TRUE)) {
+    $destination_file_scheme = file_uri_scheme($file->destination);
+    if ($destination_file_scheme !== "private" && user_is_anonymous()) {
+      $errors[] = "You don't have permissions to upload files to this folder.";
```
gillesdeudon: I suggest to mention that if the problem persists, the user must contact the administrator.

I think also that we should log the error (watchdog).

gillesdeudon (changes requested):

Fefaine:
PR added to the ticket. Currently blocked

PR #1400, opened 2017-04-28
NEPT-970: Update Europa Component rows style plugin.

drupol:
~~There's one issue with this plugin and I can't afford to spend more time in it, I almost sure this comes from the Drupal's form handling and I'm not able to find a working solution.~~

~~The problem is in Views. When you select the row plugin "Europa Component".~~
~~When you edit the settings of the plugin, you have to select which hook theme to use.~~
~~Then, once it's selected, the form reloads in AJAX and shows you the mapping between the variables of the theme hook and the fields available in the view.~~

~~Once you do your mapping and hit APPLY, for a reason that I do not know, the popup form will lose its settings.~~
~~Here, the mapping will be lost and you will have to redo it once again.~~

~~Another point of discussion in this module is, should we include ALL the hook theme in the SELECT options or only some specifics ones ?~~

~~By default, the theme hooks are gathered when we flush the caches. I know it's ugly, but as I said in the comment:~~

> AFAIK, there is no other way to do this horrendous hack to retrieve ALL the themes hooks available.
> We could use the regular way to do it with module_invoke_all('theme') and it works properly. The only issue is that theme hooks defined in the theme are not detected because we are using an admin theme. Any alternative way is very welcome.

Anyway, I'm waiting for feedback regarding these points then I'll continue my work on it.


drupol:
I think we should move the filter from the plugin to the hook_flush_caches(), what do you think @Fefaine  ?

netlooker:
During investigation of the ajax issue I've noticed that maybe it's good to align folder structure naming conventions and good practices for the Views plugins.
Most of the contrib modules stores plugins in the `views` folder. My proposal is something like this
> views/plugins/nexteuropa_core_views_plugin_row_component_view.inc

drupol:
Doing it right now, thanks for the feedback @netlooker !

drupol:
Shouldn't we use registry_autoload[] = PSR-4 in the info file ?

netlooker:
Performing final checks.

Fefaine:
Good grid implementation, i will do the same for the expandable.
In the meantime, please see my comment in the Jira ticket, i reopened it. Thanks!

profiles/common/modules/custom/nexteuropa_core_views/nexteuropa_core_views.info
```diff
@@ -3,6 +3,5 @@ description = This modules provides views plugins to implement ECL components.
 package = NextEuropa
 core = 7.x
 
-files[] = plugins/card/views_europa_card_plugin_rows.inc
-files[] = plugins/banner/views_europa_banner_plugin_rows.inc
-files[] = plugins/timelines/views_europa_timelines_plugin_rows.inc
+registry_autoload[] = PSR-4
+dependencies[] = drupal:views
```
drupol: Views is not a submodule of Drupal so this is not valid.

We use this notation when we want to enable a submodule of a module.

Exemple:

`admin_menu:admin_menu_toolbar`

profiles/common/modules/custom/nexteuropa_core_views/nexteuropa_core_views.module
```diff
@@ -11,3 +10,82 @@
...
+function _nexteuropa_core_views_row_styles() {
+  $ne_core_views_row_styles = [
+    'banner',
+    'card',
+    'blockquote',
+    'timelines',
+    'expandable',
+  ];
+  // Calling all modules implementing hook_my_data_alter():
+  drupal_alter('ne_core_views_row_styles', $ne_core_views_row_styles);
```
drupol: First, I think we shouldn't shortcut the name nexteuropa to ne.

Second, I would create a new hook: `nexteuropa_core_views_row_styles_info()`. This hook will return the array list, just like you do now.

Then, in `nexteuropa_core_views_theme_registry_alter()`, I'd call a [module_invoke](https://api.drupal.org/api/drupal/includes%21module.inc/function/module_invoke/7.x) on that hook to retrieve them all.

I think it's cleaner and more "Drupalish" :-)

drupol:
Hi,

We should also include a .api.php file on the root of the module.

Documentation: https://www.drupal.org/docs/develop/coding-standards/api-documentation-and-comment-standards

Example: https://api.drupal.org/api/drupal/modules%21system%21system.api.php/7.x

drupol:
Thanks for the changes @imanoleguskiza 👍

imanoleguskiza:
👍

voidtek:
:+1:

imanoleguskiza:
👍

PR #1410, opened 2017-05-12
NEPT-903: Removed ec_resp usage

imanoleguskiza:


drupol:
Hello,

I had doubts with the ternary operators, so I did a test: https://eval.in/798455
Evaluating a string or a int, if it's zero, it's considered as false.

This is one of the reason I don't like PHP. Anyway...

The patch seems good, but let's take the worst case and the variable 'theme_default' is not set... what do we do ?
Shouldn't we provide a theme fallback when the variable is not set ? Bartik for example ?

imanoleguskiza:
I talked about it with JF and he is the one that proposed removing it completely. I don't dislike the idea of adding a failsafe value like bartik myself, so I will let JF decide this one.

imanoleguskiza:
We discussed during todays scrum and the team decided to keep the default value empty.

Fefaine:
:+1:

drupol:
👍

PR #1420, opened 2017-05-30
NEPT-619: Alternate template

imanoleguskiza:
### Description

 [Insert description here]

### Additional commands

```sh
[Insert commands here]

```


voidtek:
Template modified in nept-619

PR #1443, opened 2017-06-09
NEPT-1163: Removing the base path from the purge paths

netlooker:
Quick fix to align with the trimmed purge paths which are expected by Varnish.

### DIFF
--- a/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
+++ b/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
@@ -506,18 +506,7 @@ function _nexteuropa_varnish_trim_base_path($path) {
   $base_path_position = strpos($path, $base_path);
 
   if ($base_path_position !== FALSE) {
-    $trimmed_path = substr_replace($path, '', $base_path_position, strlen($base_path));
-
-    // Getting the first character of the path. Function mb_substr is used
-    // because of the UTF-8 multibyte encoding.
-    $first_char = mb_substr($trimmed_path, 0, 1, 'utf-8');
-
-    // Checking if the first character of the path is not a '/'.
-    if ($first_char != "/") {
-      $trimmed_path = "/" . $trimmed_path;
-    }
-
-    return $trimmed_path;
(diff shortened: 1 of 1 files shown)


profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -486,3 +491,25 @@ function _nexteuropa_varnish_temporary_message() {
...
+ *   Trimmed path.
+ */
+function _nexteuropa_varnish_trim_base_path($path) {
+  $base_path = base_path();
+  $base_path_position = strpos($path, $base_path);
+
+  if ($base_path_position !== FALSE) {
+    $trimmed_path = substr_replace($path, '', $base_path_position, strlen($base_path));
+
+    return "/" . $trimmed_path;
```
jfhovinne: I would only add the leading '/' if it is not present yet; testing value before modifying it is required here IMO.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -486,3 +491,34 @@ function _nexteuropa_varnish_temporary_message() {
...
+  if ($base_path_position !== FALSE) {
+    $trimmed_path = substr_replace($path, '', $base_path_position, strlen($base_path));
+
+    // Getting the first character of the path. Function mb_substr is used
+    // because of the UTF-8 multibyte encoding.
+    $first_char = mb_substr($trimmed_path, 0, 1, 'utf-8');
+
+    // Checking if the first character of the path is not a '/'.
+    if ($first_char != "/") {
+      $trimmed_path = "/" . $trimmed_path;
```
netlooker: I've added the slash to keep the previous implementation. Please check the '_nexteuropa_varnish_prepare_request_options' function which is preparing the paths before they are send.

netlooker: After short discussion we've decided to remove the additional '/' part.

gillesdeudon:
:+1:

jfhovinne:
:+1:

PR #1458, opened 2017-06-21
NEPT-1185: Drupal core security update.

drupol:
### Description

Security update of Drupal core. See [changelog](https://www.drupal.org/project/drupal/releases/7.56).


drupol:
Some patches needs to be adapted.

PR #1467, opened 2017-06-27
NEPT-1031: add the theme_dropdown function and change the logic for link sending

voidtek:
### Description

Add the theme_dropdown function and change the logic for link sending

### Change log

- Changed: Add the theme_dropdown function and change the logic for link sending

### Commands

```sh
drush cc all

```


profiles/common/modules/custom/multisite_create_button/multisite_create_button.module
```diff
@@ -42,7 +42,7 @@ function _multisite_create_button_contextual_links() {
   if (!empty($content)) {
     $links = array();
     foreach ($content as $content_type) {
-      $links['items'][] = l($content_type['link_title'], $base_url . '/' . $content_type['link_path']);
+      $links['items'][$content_type['link_title']] = $base_url . '/' . $content_type['link_path'];
     }
 
     $output = theme('dropdown', $links);
```
drupol: Instead of using a `theme()` function in a `theme()` function, why not removing the `theme_dropdown` and use theme('link') directly ?

voidtek: We do not have the same logic into function between the themes. 
The theme_dropdown() in the module define the default (good practice when you declare a new theme pattern)
In the other themes, you can see the functions that override this theme_dropdown() according with the tpl.
ec_resp: https://github.com/ec-europa/ec_resp/pull/3/files#diff-13e5580ca26ea5de947e855a28fb0774R1563
ec-europa: https://github.com/ec-europa/ec-europa-theme/pull/178/files#diff-13e5580ca26ea5de947e855a28fb0774R108

imanoleguskiza:
👍

Fefaine:
👍

PR #1475, opened 2017-06-29
NEPT-873: Integrate cart with Poetry

dxvargas:
### Description

Adapt Cart and integrate with Poetry

### Change log

- Added:
Behat tests for cart
- Changed:
TBD

### Commands

```sh
[Insert commands here]

```


profiles/common/modules/features/nexteuropa_dgt_connector/nexteuropa_dgt_connector.module
```diff
@@ -3,3 +3,59 @@
...
+ * @param object $entity
+ *    The entity to translate.
+ *
+ * @return int
+ *    Length of all translatable content.
+ */
+function _nexteuropa_dgt_connector_get_content_length($entity) {
+
+  $length = 0;
+  $instances = field_info_instances($entity->entity_type, $entity->type);
```
netlooker: Here I would suggest using not very well documented but useful function.
```
$fields = field_read_fields(array(
  'entity_type' => 'node',
  'bundle' => 'page',
  'module' => 'text',
  'translatable' => TRUE,
));
```
In that way you will have all of the fields in one shot.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -382,14 +370,38 @@ function _tmgmt_poetry_form_tmgmt_job_submit($form, &$form_state) {
...
   // Find unchecked translations to be removed.
-  $result = _tmgmt_poetry_obtain_related_translation_jobs($language_targets_to_remove, 'SUB_' . $origin_job->tjid);
-  // If any, removed found translations.
-  if ($result) {
-    foreach ($result as $row) {
-      // Remove if the target language is the same as the main translation.
-      if ($origin_job->tjid !== $row->tjid && $origin_job_item->tjiid !== $row->tjiid) {
-        entity_get_controller('tmgmt_job')->delete(array($row->tjid));
-        entity_get_controller('tmgmt_job_item')->delete(array($row->tjiid));
+  // This is usefull for the #cart (but it was here before).
```
dxvargas: True, but what we had here before didn't make any sense (it was done by @imanoleguskiza long time ago and I checked with him). I took a lot of time to understand this, I had to, now it is a good opportunity for improvement. Like this, we are improving the existing code instead of leaving the nonsense here and adding more code anywhere else.
Of course we have a small risk (I say small because tests are passing still), but it's worth to take it.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
 /**
  * Implements hook_form_FORM_ID_alter().
  *
+ * Change the #cart form to provide the tmgmt_poetry module functionalities.
```
dxvargas: I've removed the description and I've added a comment inside.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
...
+function tmgmt_poetry_form_tmgmt_ui_cart_content_alter(&$form, &$form_state) {
+  $form['request_translation']['#submit'] = ['_tmgmt_poetry_workbench_store_request_languages_callback'];
+}
+
+/**
+ * Change the #cart submit logic to provide tmgmt_poetry module functionalities.
+ *
+ * @see tmgmt_ui_cart_request_translation_form_submit()
+ */
+function _tmgmt_poetry_workbench_store_request_languages_callback(&$form, &$form_state) {
```
dxvargas: It's because this function is replacing the original "tmgmt_ui_cart_request_translation_form_submit".
I just want to change the minimal code here, so that it can be better compared, what we are tweaking.

But I will put this code into the new module tmgmt_dgt_connector, it's all about the cart/small jobs.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
...
+  // Loop over all target languages, create a job for each source and target
+  // language combination add add the relevant job items to it.
+  foreach ($target_languages as $target_language) {
+    foreach ($job_items_by_source_language as $source_language => $job_items) {
+      // Skip in case the source language is the same as the target language.
+      if ($source_language == $target_language) {
+        continue;
+      }
+
+      if (isset($reference)) {
```
dxvargas: We are inside loops here. First time it's always false but later it can be set, and become true for next loops.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
...
+          // the cart.
+          $job->addItem($job_item->plugin, $job_item->item_type, $job_item->item_id);
+          $remove_job_item_ids[$job_item->tjiid] = $job_item->tjiid;
+          $job_empty = FALSE;
+        }
+        catch (Exception $e) {
+          // If an item fails for one target language, then it is also going
+          // to fail for others, so remove it from the array.
+          unset($job_items_by_source_language[$source_language][$id]);
+          drupal_set_message($e->getMessage(), 'error');
```
dxvargas: Same again, this function is replacing the original "tmgmt_ui_cart_request_translation_form_submit".
This chunk of code is coming from the original function.
So I slightly disagree that we should change the behavior here.
May we leave the decision for the second reviewer?

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
...
+        }
+      }
+
+      if (!$job_empty) {
```
dxvargas: Same again, this function is replacing the original "tmgmt_ui_cart_request_translation_form_submit".
This chunk of code is coming from the original function.
So I slightly disagree that we should change the behavior here.
May we leave the decision for the second reviewer?

I did add a comment for next lines, when setting the $reference. Here it's our code, so here sure, I agree.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -738,6 +750,121 @@ function tmgmt_poetry_page_alter(&$page) {
...
+        $jobs[] = $job;
+
+        if (!isset($reference)) {
```
dxvargas: Same again, we are inside loops, so first time it will be true, then always false. And it's used above in next loops.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
netlooker: Could you show me where in the scope of this function you are assigning a value to it?

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/controller/tmgmt_poetry.controller.job.inc
```diff
@@ -44,9 +44,15 @@ class TMGMTPoetryJobController extends TMGMTJobController {
             continue;
           }
 
-          // Check if job is already created
-          // and in that case don't create it again.
-          if (_tmgmt_poetry_obtain_related_translation_jobs(array($requested_language), 'SUB_' . $origin_job->tjid)->rowCount() > 0) {
+          // Check if job is already created and update it,
```
netlooker: Why are we forced to use tmgmt_poetry? We should have a separate implementation in the separated feature.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.plugin.inc
```diff
@@ -80,12 +80,8 @@ class TMGMTPoetryTranslatorPluginController extends TMGMTDefaultTranslatorPlugin
    * and the number of items.
    */
   public function canTranslate(TMGMTTranslator $translator, TMGMTJob $job) {
-    // Check that only one item is being requested.
-    $items = $job->getItems();
-    if (count($items) > 1) {
-      return FALSE;
-    }
     // Check that the item does not have pending requests.
+    $items = $job->getItems();
```
netlooker: In my opinion, we should implement a new TMGMTDefaultTranslatorPlugin for the 'Small Jobs' in the separated feature.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -284,34 +284,47 @@ function _tmgmt_poetry_form_load_translator_settings(&$form_state, &$form, $job)
  * Job Form Callback.
  */
 function _tmgmt_poetry_form_tmgmt_job_submit($form, &$form_state) {
+  // Current implementation part.
```
netlooker: In my opinion, we shouldn't change anything in the tmgmt_module and provide all overrides in a separated feature.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.plugin.inc
dxvargas: Yes, it's a temporary change which will be reverted in NEPT-1117, because only than we'll have the new translator. All changes are going into a feature branch, so it won't have an impact in main branches. So, I ask for the acceptance of this temporary change here, pledging it will be reverted.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
dxvargas: As I mentioned in other comments, changes here are an improvement to existing code (see comment below) and a necessary extension to work with small jobs without duplicating too much code (see comment above).

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.plugin.inc
netlooker: Ok, make sense as a temporary solution.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/controller/tmgmt_poetry.controller.job.inc
netlooker: Here I don't fully agree. Duplication is a lot cheaper than bundling two implementations in one place. At least it's my opinion. Let's give it a final word to the second reviewer.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
netlooker: Ok, so let's give a chance to speak to another reviewer.

netlooker:
I've approved the changes but here to solve some concerns and questions we need to have another pair of eyes. Let's go for code review +1.

PR #1483, opened 2017-07-07
NEPT-1209: Revert job items policy, create them along with job

dxvargas:
### Description

Revert job items policy, create them along with job.

### Change log

- Removed: Logic to delay creation of Job Items only when submitting the translation request. It should use the default behavior of TMGMT, which is to create Job and Job Items all together.

### Commands

No extra commands are needed.


gillesdeudon:
👍

Fefaine:
👍

PR #1484, opened 2017-07-10
NEPT-1172: Testing Behat implementation on 2.4 codes

gillesdeudon:
### Description

Temporary branch  that is just there for test.

**It should be destroyed after the tests and never merged in the master branch!**


gillesdeudon:
I close the PR as this branch has fulfil its goal and must not be merge in master.

PR #1490, opened 2017-07-13
NEPT-5: Fix security issue in communities module

ccontet:
### Description

Fix security issue in communities module

### Change log

- Added:
- Changed: SQL request
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
no command

```


Fefaine:
The ticket nr is not correct. The ticket nr must be correct because the PR message is used to generate the changelog.

Fefaine:
👎

PR #1510, opened 2017-07-31
NEPT-1234: Remove the field_custom_js_file file into nexteuropa_webtools.

voidtek:
### Description

Remove the field_custom_js_file file into nexteuropa_webtools.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed: field_custom_js_file file into nexteuropa_webtools.
- Fixed:
- Security:

### Commands

```sh
drush cc all

```


Fefaine:
outdated, we'll create a new one

PR #1524, opened 2017-08-17
NEPT-1429: Update variable for platform version

dxvargas:
### Description

Update variable for platform version.

### Change log

- Added: hook_update_N in cce_basic_config to update the variable for platform version.

### Commands

```sh
drush updb
```


voidtek:
The code is good.

PR #1539, opened 2017-08-25
NEPT-1286: Removed redeclaration of global variable.

imanoleguskiza:
### Description

Removed redeclaration of global variable.

### Change log

- Fixed: emoved redeclaration of global variable.


gillesdeudon:
Ok for me too

PR #1541, opened 2017-08-25
NEPT-1287: Removed empty lines after inline comments.

imanoleguskiza:
### Description

Removed empty lines after inline comments.

### Change log
- Fixed: NEPT-1287: Removed empty lines after inline comments.


voidtek:
👍

PR #1542, opened 2017-08-25
NEPT-1094: Define the component social_icon.

voidtek:
### Description

Define the component social_icon into nexteuropa_formatters

### Change log

- Added: the definition of the component social_icon.

### Commands

```sh
no commands

```

### DIFF
--- a/profiles/common/modules/custom/nexteuropa_formatters/nexteuropa_formatters_fields/nexteuropa_formatters_fields.module
+++ b/profiles/common/modules/custom/nexteuropa_formatters/nexteuropa_formatters_fields/nexteuropa_formatters_fields.module
@@ -57,6 +57,26 @@ function nexteuropa_formatters_fields_field_formatter_info() {
         'show_translations' => TRUE,
       ),
     ),
+    'ne_social_icon' => array(
+      'label' => t('NE Social Icon'),
+      'description' => t('Displays as a Social Icon'),
+      'field types' => array(
+        'link_field',
+      ),
+      'settings' => array(
+        'type' => 'email',
+      ),
+    ),
+    'ne_social_icon_link' => array(
+      'label' => t('NE Social Icon - link'),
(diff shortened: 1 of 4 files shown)


profiles/common/modules/custom/nexteuropa_formatters/nexteuropa_formatters_fields/nexteuropa_formatters_fields.module
```diff
@@ -57,6 +57,26 @@ function nexteuropa_formatters_fields_field_formatter_info() {
...
       ),
     ),
+    'ne_social_icon' => array(
+      'label' => t('NE Social Icon'),
+      'description' => t('Displays as a Social Icon'),
+      'field types' => array(
+        'link_field',
+      ),
+      'settings' => array(
+        'type' => 'email',
```
Fefaine: I don't understand why you are using an email type?

drupol: This is the formatter's default config.

drupol: See: https://api.drupal.org/api/drupal/modules%21field%21field.api.php/function/hook_field_formatter_info/7.x

voidtek: I confirm, it's the default value for this type of component.
I don't want put a social network (like facebook, twitter or other) as a default, so I choose one value that exists and that every body knows. 
We can try to put with no default value.

drupol: 👍

netlooker:
👍

PR #1552, opened 2017-08-28
NEPT-1299: Removed db_select calls.

imanoleguskiza:
### Description

Removed db_select calls.

### Change log

- Fixed: Removed db_select calls.


voidtek:
:+1:

profiles/common/modules/custom/multisite_config/lib/Drupal/locale/Config.php
```diff
@@ -61,8 +61,12 @@ public function setLanguageNegotiation($language_negotiation, $type = LANGUAGE_T
    *    Language prefix.
    */
   public function setLanguagePrefix($language, $prefix) {
-    $arguments = array(':language' => $language, ':prefix' => $prefix);
-    db_query("UPDATE {languages} SET prefix = ':prefix' WHERE language = ':language'", $arguments)->execute();
+    db_update('languages')
+      ->fields(array(
+        'prefix' => $prefix,
+      ))
+      ->condition('language', $language, '=')
```
Fefaine: imo, the condition is more clear without the operator

> If no operator is specified, = is assumed. That means the most common case would be condition('myfield', $value), which results in a conditional fragment of myfield = :value, where :value will be replaced with $value when the query is run.

If you agree i will make the small change

profiles/common/modules/custom/multisite_og_navigation_tree/multisite_og_navigation_tree.install
```diff
@@ -8,7 +8,12 @@
  * Implements hook_install().
  */
 function multisite_og_navigation_tree_install() {
-  db_query("UPDATE {system} SET weight = 5 WHERE name = 'multisite_og_navigation_tree'");
+  db_update('system')
+    ->fields(array(
+      'weight' => 5,
+    ))
+    ->condition('name', 'multisite_og_navigation_tree', '=')
```
Fefaine: idem

Fefaine:
The change is just a detail, ready to merge.

PR #1554, opened 2017-08-28
NEPT-1097: Fix regression caused by the renaming of the module formatters

Fefaine:
### Description

Fix regression caused by the renaming of the module formatters

### Change log

- Fixed: hook name


drupol:
:+1:

PR #1555, opened 2017-08-29
NEPT-1453: Fix text format filters

gillesdeudon:
### Description

Add the "Sanitize HTML" filter to full html formats and remove a permission at installation time.

### Change log

- Changed: Sanitize html filter to formats.
- Changed: Change format permissions.

### Commands

The changes are only for fresh install and does not imply the existing site.


gillesdeudon:
I close the ticket because of Devop instruction.

PR #1560, opened 2017-08-29
NEPT-1306: Locked all fields.

imanoleguskiza:
### Description

Locked all fields, including body field.

### Change log

-Fixed: Locked all fields.


Fefaine:
👎

PR #1569, opened 2017-08-30
NEPT-1364: Added exclusion rule until standard is defined.

imanoleguskiza:
## NEPT-1364

Added exclusion rule until standard is defined.

### Change log
- Fixed: Added exclusion rule until standard is defined.


Fefaine:
👍

PR #1573, opened 2017-08-30
NEPT-1368: Removed extra space on return comments.

imanoleguskiza:
### Description

Removed extra space on return comments.

### Change log

- Fixed: Removed extra space on return comments.


Fefaine:
Newsroom needs to be excluded.

Fefaine:
Ok for me. Waiting build completion before to merge.

PR #1574, opened 2017-08-30
NEPT-1369: Aligned equal signs.

imanoleguskiza:
### Description
Aligned equal signs.

### Change log

- Fixed: Aligned equal signs.


Fefaine:
Ok for me

PR #1578, opened 2017-08-30
Nept 1440: Removal of the feature likedislike

Fefaine:
### Description

Remove outdated feature likedislike

### Change log

- Removed: Feature like dislike and symlinks in profile

### Commands

```sh
no commands. 
```


profiles/common/modules/custom/multisite_cookie_consent/multisite_cookie_consent.module
```diff
@@ -50,8 +50,7 @@ function multisite_cookie_consent_init() {
   if (current_path() != 'cookie-notice') {
     // Notice url.
     $notice_page = url('cookie-notice', array('absolute' => TRUE));
-    // Add for example external.
-    drupal_add_js('http://ec.europa.eu/wel/like/js/likedislike.js', 'external');
```
voidtek: We cannot remove this line, we should change the link.
for example remplace by http://europa.eu/webtools/load.js

Fefaine: Why do you want to replace this line? This feature no longer exists, so the js is no longer needed.

Fefaine: Module was removed from the platform

imanoleguskiza:
👍

PR #1579, opened 2017-08-30
NEPT-672: Upgrading file_entity module to latest version.

Fefaine:
### Description

Upgrade file_entity module to version 2.4 and add the patch on d.o issue 2893132

### Change log

- Changed: file_entity module version

### Commands

```sh
no command
```

### DIFF
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -288,11 +288,10 @@ projects[field_group][version] = "1.5"
 ; https://webgate.ec.europa.eu/CITnet/jira/browse/NEXTEUROPA-6603
 projects[field_group][patch][] = https://www.drupal.org/files/issues/field_group_label_translation_patch.patch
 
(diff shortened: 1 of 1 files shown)


resources/multisite_drupal_standard.make
```diff
@@ -289,7 +289,8 @@ projects[field_group][version] = "1.5"
 projects[field_group][patch][] = https://www.drupal.org/files/issues/field_group_label_translation_patch.patch
 
 projects[file_entity][subdir] = "contrib"
-projects[file_entity][version] = "2.2"
+projects[file_entity][version] = "2.4"
+projects[file_entity][patch][] = https://www.drupal.org/files/issues/D7-file_entity-file_description_missing-2893132-2.patch
```
voidtek: Can you add information of this patch?

 ; Issue #99999: blablablabla.
 ; https://www.drupal.org/node/99999
 ; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-999

voidtek: If you want to upgrade the module for the last version, maybe we can point directly the last commit.

PR #1582, opened 2017-08-31
NEPT-1289: Remove the cs rule on a feature that is no longer part of the platform.

Fefaine:
Remove the cs rule on a feature that is no longer part of the platform.

### Description

Remove the cs rule on a feature nexteuropa_subscriptions

### Change log

- Removed: cs exception on nexteuropa_subscriptions

### Commands

```sh
no commands
```


drupol:
:+1: all good

PR #1584, opened 2017-09-04
NEPT-1412: Fix QA automation rules - There must be one blank line after the last USE statement

Fefaine:
### Description

There must be one blank line after the last USE statement

### Change log

- Changed: line spacing after use statement

### Commands

```sh
no command

```


dxvargas:
👍

PR #1594, opened 2017-09-06
NEPT-1309: Added temporary exclusion rule

voidtek:
### Description

Added temporary exclusion rule

### Change log

- Added: temporary exclusion rule

### Commands

```sh
no commands

```


dxvargas:
👍

PR #1605, opened 2017-09-19
NEPT-1255: Remove the check_plain from the ecas_import_users.

Fefaine:
### Description

Remove the check_plain from the ecas_import_users.

### Change log

- Removed: Unnecessary check_plain

### Commands

```sh
no command
```


dxvargas:
:+1:

PR #1608, opened 2017-09-22
NEPT-1516: Avoid changes to file in updb when it is already translatable

dxvargas:
### Description

Shuffle NEPT-7 code to not create regressions when updating subsites (original File fields values not being fetched).

### Change log

- Changed: Check if Files are already translatable, if so skip changes;
- Changed: When defining Files to be translatable, set default language to LANGUAGE_NONE;
- Removed: Hook update `cce_basic_config_update_7217` (merged into `nexteuropa_multilingual_update_7002`);
- Fixed: Dependent File fields were loosing their values.


voidtek:
👍

PR #1611, opened 2017-09-25
NEPT-1540: Fix nexteuropa_varnish problem with the subsites having "/…

Fefaine:
…" as base URL.

### Description

First slash is removed from purge rule.

### Change log

- Changed: _nexteuropa_varnish_trim_base_path

### Commands

```sh
no command to run
```


jfhovinne:
:-1: I would suggest to add tests since it seems this edge case is not covered yet.

gillesdeudon:
👍

PR #1619, opened 2017-09-27
NEPT-1519: Improve DGT Translator

dxvargas:
### Improve DGT Translator

Align translator with latest improvements in Poetry Connector and NEPT-880 conclusions.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/tmgmt_dgt_connector.module
```diff
@@ -82,13 +101,10 @@ function _tmgmt_dgt_connector_workbench_store_request_languages_callback(&$form,
   global $user;
 
   $target_languages = array_filter($form_state['values']['target_language']);
-  $enforced_source_language = NULL;
-  if ($form_state['values']['enforced_source_language']) {
-    $enforced_source_language = $form_state['values']['source_language'];
-  }
-
+  $enforced_source_language = $form_state['values']['enforced_source_language'] ? $enforced_source_language = $form_state['values']['source_language'] : NULL;
```
imanoleguskiza: This statement looks weird, shouldn't it be $enforced_source_language = $form_state['values']['enforced_source_language'] ? $form_state['values']['source_language'] : NULL;

tests/behat.yml.dist
```diff
@@ -48,7 +48,7 @@ default:
     EC\Behat\PoetryExtension:
       application:
         base_url: '{{ behat.base_url }}'
-        endpoint: '/dgt_ftt_translator/dgt_service_callback'
+        endpoint: '/tmgmt_dgt_connector/server'
```
imanoleguskiza: We need to keep and eye on this change and revert it (or change it) when we finish https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1601

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/tmgmt_dgt_connector.info
```diff
@@ -5,3 +5,6 @@ core = 7.x
 dependencies[] = tmgmt_poetry
 files[] = plugins/tmgmt_dgt_connector.plugin.inc
 files[] = plugins/tmgmt_dgt_connector.ui.inc
+
+dependencies[] = registry_autoload
+registry_autoload[] = PSR-4
```
imanoleguskiza: Again the empty lines :S

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/src/Subscriber.php
```diff
@@ -0,0 +1,372 @@
...
+   * Listener for the event onTranslationReceivedEvent.
+   *
+   * @param \EC\Poetry\Events\Notifications\TranslationReceivedEvent $event
+   *   The event for the Translation Received.
+   *
+   * @return bool
+   *   Return True if the translation is received without issues.
+   */
+  public function onTranslationReceivedEvent(TranslationReceivedEvent $event) {
+    $translator = tmgmt_translator_load(TMGMT_DGT_CONNECTOR_TRANSLATOR_NAME);
```
imanoleguskiza: I think both onTranslationReceivedEvent and onStatusUpdatedEvent should check if the translator associated to a job is the tmgmt_dgt_connector one or not. This will prevent it from triggering on other translators (like Fast Track)

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/plugins/tmgmt_dgt_connector.ui.inc
```diff
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * Provides Google Translator ui controller.
+ * Provides Translator ui controller.
```
imanoleguskiza: Why not add a DGT to the string :P

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/src/Subscriber.php
dxvargas: You're right @imanoleguskiza . This is the same situation we faced in Fast Track.
But I propose to postpone that change to a new ticket because of two strong reasons:
- This ticket/PR has consumed already a huge amount of time and, as pledged in the retrospective meeting, we should create more tickets with precise scope (this is the same approach we had for Fast Track, where a special ticket was created, [NEPT-1590](https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1590));
- Within [NEPT-1601](https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1601) we are planning to handle this logic in a proper way, so that we don't need to do the same steps (check reference, get job, get translator, then proceed or skip the request) in each listener.

PR #1625, opened 2017-09-29
NEPT-392: Upgrade i18n, remove patch.

Fefaine:
### Description

Upgrade i18n module, remove a patch.

### Change log

- Changed: i18n module version

### Commands

```sh
no command
```


gillesdeudon:
OK

PR #1629, opened 2017-10-03
NEPT-1385 Namespaced classes/interfaces/traits with "use" statements

aritomelo:
### Description
   Fix QA automation rules - Namespaced classes/interfaces/traits should be referenced with use statements

### Change log

- Removed: The rule ref="Drupal.Classes.FullyQualifiedNamespace.UseStatementMissing"

- Fixed: Include the classes by using USE and namespacing

### Commands

NA.


dxvargas:
👍

PR #1633, opened 2017-10-03
NEPT-1386 When importing a class with "use", do not include

aritomelo:
### Description

Fix QA automation rules - When importing a class with "use", do not include a leading backslash

   ### Change log
- Removed: The rule ref="Drupal.Classes.UseLeadingBackslash.SeparatorStart"
- Removed: All back slash fire up by phpcs

### Commands

NA.


Medi4tor:
Seems good for me.

dxvargas:
:+1

PR #1639, opened 2017-10-04
NEPT-1389 Return type "int | NULL" must not contain spaces

aritomelo:
### Description
Fix QA automation rules - Return type "int | NULL" must not contain spaces.

### Change log
  - Removed: The rule ref="Drupal.Commenting.FunctionComment.ReturnTypeSpaces"    
    
  - Removed: The unnecessary spaces.

  - Created:  An exception has been set to exclude nexteuropa_newsroom module once it is not 
                    supported by Core team, so to progress with the clean up for QA AUTOMATION (NEPT-
                    884, NEPT-1387 and NEPT-1389), this exception must exist so when remove the          
                    "Drupal.Commenting.FunctionComment.ParamMissingDefinition"  rule the phpcs is OK 
                    (all green).

### Commands

NA.


Medi4tor:
Ok for me.

gillesdeudon:
👍

PR #1642, opened 2017-10-05
NEPT-1390 Fix QA automation rules - Line indented incorrectly

aritomelo:
### Description

Fix QA automation rules - Line indented incorrectly.

### Change log
  - Removed: The rule ref="Drupal.WhiteSpace.ScopeIndent.IncorrectExact"

  - Added: The right indentation.

  - Created:  An exception has been set to exclude nexteuropa_newsroom module once it is not 
                    supported by Core team, so to progress with the clean up for QA AUTOMATION (NEPT-
                    884, NEPT-1387 and NEPT-1389), this exception must exist so when remove the rule   
                    "Drupal.Commenting.FunctionComment.ParamMissingDefinition" the phpcs is OK (all 
                     green).

### Commands

NA.

### DIFF
--- a/phpcs-qa-roadmap.xml
+++ b/phpcs-qa-roadmap.xml
@@ -608,11 +608,6 @@
         <exclude-pattern>*</exclude-pattern>
     </rule>
 
-    <!-- Line indented incorrectly. -->
-    <rule ref="Drupal.WhiteSpace.ScopeIndent.IncorrectExact">
-        <exclude-pattern>*</exclude-pattern>
-    </rule>
-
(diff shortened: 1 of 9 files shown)


Medi4tor:
Hi @aritomelo , was it really necessary to add this new exception rule?
as we intent to remove exception rules.
Do we know who is supporting this module? Do we still need this module in the platform?

Medi4tor:
@aritomelo 
Seems like the branch number does not match the pull request ticket id.
Can you check?

Medi4tor:
Seems good for me.

gillesdeudon:
👍

PR #1644, opened 2017-10-05
NEPT-1394 Do not concatenate strings to translatable strings, use placeholders

aritomelo:
### Description
     Fix QA automation rules - Do not concatenate strings to translatable strings, they should be part of the t() argument and you should use placeholders.

  ### Change log
  - Removed: The rule
```java
    <!-- Do not concatenate strings to translatable strings, they should be part of the t() argument and you should use placeholders. -->
    <rule ref="Drupal.Semantics.FunctionT.ConcatString">
        <exclude-pattern>*</exclude-pattern>
    </rule>

```

    - Created:  An exception has been set to exclude nexteuropa_newsroom module once it is not supported   by Core team, so to progress with the clean up for
     QA AUTOMATION (NEPT-884, NEPT-1387, NEPT-1389, NEPT-1390, NEPT-1391), this exception must exist
     so when removing the rule the phpcs is OK (all green).

  ### Commands
```sh
    git fetch --all
    git checkout nept-1394
    ./bin/phpcs

```


Medi4tor:
Ok for me.

gillesdeudon:
👍

PR #1647, opened 2017-10-06
NEPT-528: Remove inject_data from communities profile.

Fefaine:
### Description
Remove inject_data from communities profile.

### Change log

- Removed: outdated file name inject_data.php

### Commands

```sh
no commands
```


gillesdeudon:
OK

PR #1650, opened 2017-10-06
NEPT-1176: Drupal logout malfunction

gillesdeudon:
### Description

Fix the redirection process that stopped directly after the EU Login process.

### Change log

- Fixed: the redirection definition in the logout call in the ECAS module.

### Commands

```sh
No command required

```


dxvargas:
👍

PR #1653, opened 2017-10-06
NEPT-1398 Values usually have to run through t() for translation

aritomelo:
## [NEPT-1398](https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1398)

  ### Description 
Fix QA automation rules - Values usually have to run through t() for translation.

  ### Change log
  - Removed: The rule ref="DrupalPractice.General.OptionsT.TforValue"
 
 - Added: 
    1- Wrap all string within t function.

    2- An exception has been set to exclude nexteuropa_newsroom module once it is not 
             supported  by Core team, so to progress with the clean up for QA AUTOMATION 
             (NEPT-884, NEPT-1387, NEPT-1389, NEPT-1390, NEPT-1391, NEPT-1396, NEPT-1397), 
             this exception must exist while removing the rule and get OK from the phpcs (all green).

  ### Commands
NA.


Medi4tor:
Ok for me.

Medi4tor:
Ok with the changes.

PR #1657, opened 2017-10-10
NEPT-308: Remove dependencies to dblog from profiles.

Fefaine:
### Description

Remove dependencies to dblog from profiles

### Change log
- Removed: dblog dependency in profiles.

### Commands

no command


ademarco:
We check in after scenario if warnings have been added in the watchdog, that's why Behat fails I guess.

Medi4tor:
Sounds good. good idea.

PR #1660, opened 2017-10-10
NEPT-1494: Sanitize HTML filter: Remove the central configuration feature

drupol:
### Description

Sanitize HTML filter: Remove the central configuration feature & the node related parameter.

### Change log

- Removed: The central configuration feature & the node related parameter.

### Commands

```sh
no commands

```


drupol:
Merging...

PR #1662, opened 2017-10-10
NEPT-1402 Use LANGUAGE_NONE instead of 'und'

aritomelo:
### Description
Fix QA automation rules,  use LANGUAGE_NONE instead of und.

### Change log
  - Removed: The rule "DrupalPractice.General.LanguageNone.Und"

  - Added:  An exception has been set to exclude nexteuropa_newsroom module once it is not supported   by Core team, so to progress with the clean up for
     QA AUTOMATION (NEPT-884, NEPT-1387, NEPT-1389, NEPT-1390, NEPT-1391, NEPT-1396, NEPT-1400), this exception must exist while removing the rule and get OK from the phpcs (all green).

  - Changed: The 'und' key has been changed by LANGUAGE_NONE const as DRUPAL best practices recommended.

### Commands

No commands.

### DIFF
--- a/phpcs-qa-roadmap.xml
+++ b/phpcs-qa-roadmap.xml
@@ -254,22 +254,6 @@
         <exclude-pattern>profiles/common/modules/features/cce_basic_config/cce_basic_config.install</exclude-pattern>
     </rule>
 
-    <!-- Use the LANGUAGE_NONE constant. -->
-    <!-- https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-821 -->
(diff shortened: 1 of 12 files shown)


jfhovinne:
@aritomelo 
PR title should not contain special characters and should be shorter.
Please remove code from PR description and use PR pattern documented in the wiki.

Medi4tor:
Test comment FMSTEREO

Medi4tor:
Test FM

aritomelo:
@jfhovinne Hi, is it look right?

Fefaine:
👍

PR #1663, opened 2017-10-10
NEPT-1409: Fix QA-Remove exception for opening brace of JS functions,…

Medi4tor:
### Description

Fix QA automation rules - Expected 1 space before opening brace

### Change log

- Changed:
 Rule removed from phpcs-qa-roadmap.xml
 Space added on several files.

### Commands
 NA

### DIFF
--- a/phpcs-qa-roadmap.xml
+++ b/phpcs-qa-roadmap.xml
@@ -698,11 +698,6 @@
         <exclude-pattern>*</exclude-pattern>
     </rule>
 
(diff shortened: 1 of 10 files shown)


jfhovinne:
@Medi4tor 
PR title should not contain special characters and should be shorter.
'Commands' section is for DevOps, no need to specify branching strategy here.
Please use PR pattern as documented in the wiki.

PR #1666, opened 2017-10-10
NEPT-1405  Content missing for @var tag 

aritomelo:
### Description
Fix QA automation rules - Content missing for @var tag in member variable comment.

### Change log
  - Removed: The rule ref="DrupalPractice.General.LanguageNone.Und"
  
  - Added: 
            1- An exception has been set to exclude nexteuropa_newsroom module once it is not 
                supported   by Core team, so to progress with the clean up for
                QA AUTOMATION (NEPT-884, NEPT-1387, NEPT-1389, NEPT-1390, NEPT-1391, NEPT-
                1396, NEPT-1400, NEPT-1402, NEPT-1405), this exception must exist while removing the 
                rule and get OK from the phpcs (all green).
             2- The variable type for heredoc.

### Commands

NA.

### DIFF
--- a/phpcs-qa-roadmap.xml
+++ b/phpcs-qa-roadmap.xml
@@ -678,11 +678,6 @@
         <exclude-pattern>*</exclude-pattern>
     </rule>
 
-    <!-- Content missing for @var tag in member variable comment. -->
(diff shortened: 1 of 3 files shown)


Medi4tor:
Ok for me

PR #1667, opened 2017-10-10
NEPT-267: Upgrade diff module to latest commit.

Fefaine:
### Description
Upgrade diff module to latest commit.

### Change log

- Changed: Version of DIFF module.

### Commands

```sh
no command

```


gillesdeudon:
Ok for me

PR #1668, opened 2017-10-10
NEPT-1415  Empty strongarm settings are not allowed.

aritomelo:
## [NEPT-1415](https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1415)

### Description
Fix QA automation rules - Empty strongarm settings for 'pathauto_node_blog_pattern' are not allowed.

### Change log
  - Removed: 
     1- The rule "DrupalPractice.General.LanguageNone.Und"
     2-  Block of codes related to empty value for path_auto using strongarm.
       Accordingly, with QA drupal standard, we should: Avoid empty strongarm values
       Strongarm settings shouldn't be delivered when the value is empty, especially with pathauto 
       patterns. If it has '' value then the webmaster will change it on production it will make the feature 
       overridden.
       Solution: do not put empty settings to strongarm.
                      Find more at [QA- Strongarm](https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=MULTISITE&title=Features+files+requirements#strongarm)

  ### Commands

NA.


Medi4tor:
Good for me.

PR #1672, opened 2017-10-11
NEPT-1417  Fix QA automation rules - File appears to be minified and cannot be processed.

aritomelo:
## [NEPT-1417](https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1417)

  ### Description
      Fix QA automation rules - File appears to be minified and cannot be processed.

  ### Change log
  - Removed: The rule
```java
    <!-- File appears to be minified and cannot be processed. -->
    <rule ref="Internal.Tokenizer.Exception">
        <exclude-pattern>*</exclude-pattern>
    </rule>
```
    - Added: A rule to exclude nexteuropa_newsroom.

  ### Commands

    NA.

### DIFF
--- a/phpcs-qa-roadmap.xml
+++ b/phpcs-qa-roadmap.xml
@@ -738,11 +738,6 @@
         <exclude-pattern>*</exclude-pattern>
     </rule>
 
-    <!-- File appears to be minified and cannot be processed. -->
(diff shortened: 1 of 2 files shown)


Medi4tor:
Ok for me...

PR #1684, opened 2017-10-16
NEPT-1044: Internal content tokens display as text in a wysiwyg table

gillesdeudon:
### Description

Unblock the possibility to to add an internal content tokens inside a wysiwyg table and in a bullet list

### Change log

- Fixed: nexteuropa_token_ckeditor plugin.js to declare that the custom tag of internal content tokens can be included in table (specifically in a "td" tag) and a bullet list (specifically in a "li" tag).

### Commands

```sh
drush cc all
```


aritomelo:
Ready to be reviewed by +1

PR #1686, opened 2017-10-16
NEPT-289: Make Mpdf library PHP 7.0 compatibility

gillesdeudon:
### Description

Upgrade mpdf to 6.1.4. in order to make the library PHP 7 compliant

### Change log

- Changed: Upgrade mpdf to 6.1.4.

### Commands

```sh
drush cc all

```


Fefaine:
👍

PR #1696, opened 2017-10-20
NEPT-1568: Replace db calls with drupal api in multisite_config

imanoleguskiza:
### Description

Removed unnecesary user checks and commented db update methods in order to make clear the reasoning behind their use.

### Change log

- Added: Comments on db calls
- Changed: Removed unnecesary checks


ademarco:
👍

PR #1700, opened 2017-10-26
NEPT-1038: update patch to not show the bean h2 when there is no title.

Fefaine:
### Description

Update patch to not show the h2 when there is no title.

### Change log

- Changed: the function name to test emptiness.

### Commands

```sh
drush cc all
```


Medi4tor:
Review OK on 2.3

PR #1701, opened 2017-10-26
nept-1038: update patch to not show the bean h2 when there is no title.

Fefaine:
### Description

update patch to not show the bean h2 when there is no title.

### Change log

- Changed: function to check emptiness of title

### Commands

```sh
drush cc all
```


Medi4tor:
Reviewed: It's good!

PR #1703, opened 2017-10-27
Nept 1584: Force variable xmlsitemap_prefetch_aliases to 0.

Fefaine:
### Description

Force variable xmlsitemap_prefetch_aliases to 0. This functionality has a very negative impact on performances.

### Change log

- Changed: value of variable xmlsitemap_prefetch_aliases set to 0.

### Commands

```sh
drush updb
```

### DIFF
--- a/tests/features/sitemap.feature
+++ b/tests/features/sitemap.feature
@@ -10,14 +10,14 @@ Feature: Check the Sitemap
       | sitemap     |
     Given I am logged in as a user with the "administrator" role
     When I go to "/admin/config/search/xmlsitemap/settings"
-    Then "Prefetch URL aliases during sitemap generation." checkbox should not be checked
+    Then the "Prefetch URL aliases during sitemap generation." checkbox should not be checked
 
   @api
   Scenario: Administrator user can check the sitemap
     Given the module is enabled
       | modules     |
       | sitemap     |
-    Given I am logged in as a user with the "anonymous" role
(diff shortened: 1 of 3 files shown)


tests/features/sitemap.feature
```diff
@@ -4,11 +4,20 @@ Feature: Check the Sitemap
   I can publish a sitemap
 
   @api
+  Scenario: Value of the variable xmlsitemap_prefetch_aliases is zero
+    Given the module is enabled
+      | modules     |
+      | sitemap     |
+    Given I am an administrator user
```
Medi4tor: Hi,
Maybe we can use 
Given I am logged in as a user with the "administrator" role

Fefaine: Indeed, we have 2 different steps for this, we should only have one to be more consistent

Medi4tor:
Seems OK for me, little changes mad in behat tests.

aritomelo:
@Fefaine @Medi4tor I have checked the PR and I am wondering whether or not we should accept the second commit and the reason why is: I have tested both commits and both run without issue, so I do not know which one is the better.

Fefaine:
@aritomelo  indeed both these steps are correct 
"Given I am an administrator user"
and
"Given I am logged in as a user with the "administrator" role"
But it is stupid to have two steps for the same action, so we will probably remove the first one. That's why i finally used the second one :)

aritomelo:
All OK, ready to go!

PR #1704, opened 2017-10-27
NEPT-1577: Upgrade the version of EC_EUROPA.

voidtek:
### Description

Upgrade the version of EC_EUROPA to 0.0.4.

### Change log

- Changed: The version of EC_EUROPA to 0.0.4.

### Commands

```sh
no commands.

```


imanoleguskiza:
Same here, giving my +1 but we really really need to check why this is happening.

PR #1706, opened 2017-10-30
NEPT-1586: Remove temporary workaround for username rendering in nodes.

voidtek:
### Description

Remove temporary workaround for username rendering in nodes.

### Change log

- Fixed: Remove temporary workaround for username rendering in nodes.

### Commands

```sh
no commands

```

PR #1709, opened 2017-11-03
NEPT-1586: Remove temporary workaround for username rendering in nodes. (#1706)

voidtek:
### Description

Remove temporary workaround for username rendering in nodes. (#1706)
Fix regression caused by the renaming of the module formatters (#1554)

### Change log

- Added: Fix regression caused by the renaming of the module formatters
- Changed: Change the version of Atomium to 7.x-2.x

### Commands

```sh
no commands

```


netlooker:
👍

PR #1712, opened 2017-11-06
NEPT-263: Adding patch for PHP7 compatibilty to iCalcreator library.

Fefaine:
### Description

Patching iCalcreator library for PHP7 compatibility.

### Change log

- Added: Patch https://www.drupal.org/files/issues/iCalcreator-php-7-2707373-6.patch for PHP7 compatibility.

### Commands

```sh
no commands
```

### DIFF
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -898,6 +898,8 @@ libraries[iCalcreator][download][type] = "file"
 libraries[iCalcreator][download][request_type]= "get"
 libraries[iCalcreator][download][file_type] = "zip"
 libraries[iCalcreator][download][destination] = "../common/libraries"
+; Adding patch for PHP7 compatibilty on IcalCreator.class.
+; https://www.drupal.org/files/issues/iCalcreator-php-7-2707373-6.patch
 libraries[iCalcreator][patch][2707373] = https://www.drupal.org/files/issues/iCalcreator-php-7-2707373-6.patch
 
 ; imgAreaSelect 0.9.10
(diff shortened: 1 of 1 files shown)


aritomelo:
Taking into account that the applied patch follow those rules:
- Patch from d.org
- Patch has Green Flag
- Patch integrated through makefile

However, I'd suggest we follow this [guideline from QA](https://github.com/ec-europa/futurium-reference/blob/master/resources/site.make) .

Also, I notice that in the past we did not use the explicit index, in this case, it has [2707373] and throughout the file, I am seeing omitted index [], which one is the best practice here?

aritomelo:
@drupol Could you explain to me if there is any reason to apply a patch using its version as index, as used [here](https://github.com/ec-europa/platform-dev/blob/0872d4c342a6ee93194ddbee5625ef0245d223d0/resources/multisite_drupal_standard.make#L959)

The above solution is not synchronised with the [QA guideline](https://github.com/ec-europa/futurium-reference/blob/master/resources/site.make).

Thx in advance.

drupol:
Hi bruh',

It's up to you, I don't think there is a convention for that, maybe we should agree to use either `[]` or `[INDEX]`.

My way to do is to add the D.O. node id as index so I avoid patch duplicates.

voidtek:
Hi,
The name and the link for the issue on d.org are missing in the PR.
(see the example: https://github.com/ec-europa/platform-dev/pull/1712/files#diff-2b3d690e785dc60ff797b3555c6c2640L956)

aritomelo:
@voidtek Thanks for pointing it out, I have addressed it.

Regarding the array index, as we do not have a rule for it, I will accept the PR as it is, but we Should consider a common agreement in what to use!

PR #1715, opened 2017-11-06
Nept 500: uninstall multisite_og_navigation_tree.

Fefaine:
### Description

Uninstall multisite_og_navigation_tree.
Mark the module as deprecated.

### Change log

- Deprecated: multisite_og_navigation_tree.

### Commands

```sh
drush updb
```


Fefaine:
@ademarco @jfhovinne i would like to review the rules that fixes the make file on a commit https://webgate.ec.europa.eu/CITnet/confluence/display/NEXTEUROPA/NE+Core+internal+QA+rules
We have a problem when some dependencies expect version "higher than"
example : Module media_crop cannot be enabled because it depends on file_entity  (>7.x-2.0) but f9b172177f340204fbed3ad0ac2fdcfef0d42271 is available
& in this case, error messages when runing updb. It's possible using drush but not using the UI

Medi4tor:
Reviewed : OK

PR #1727, opened 2017-11-09
NEPT-482: Implement behat tests for taxonomy browser feature.

VictorBonasa:
### Description

Implement_behat_tests_for_taxonomy_browser_feature

### Change log

- Added: Added tests for taxonomy browser

### Commands

No commands needed


tests/features/taxonomy_browser.feature
```diff
@@ -0,0 +1,94 @@
...
+  In order to test ‘Taxonomy browser’ feature
+  As an administrator
+  I want to make sure ‘Taxonomy browser’ allows creating blocks with taxonomy vocabularies
+
+  Background:
+    Given I am logged in as a user with the 'administrator' role
+    When I visit "/admin/config/taxonomy_browser/settings_en"
+    When I check "edit-taxonomy-browser-vocabulary-2"
+    And  I press "Save configuration"
+    Then I should see the text "The configuration options have been saved."
```
imanoleguskiza: We don't need to add the configuration as a background. You can first run the scenario "Administrator user activates the vocabulary on taxonomy browser page and configures the block" and then create the view.

tests/features/taxonomy_browser.feature
```diff
@@ -0,0 +1,94 @@
...
+    Then I check "Content: Has taxonomy term ID (with depth)"
+    And I press the "Add and configure contextual filters" button
+    When I select the radio button "Provide default value"
+    And I fill in "Fixed value" with "fixed_value"
+    Then I check "Override title"
+    And I fill in "edit-options-title" with "%1"
+    And I press the "Apply" button
+    And the module is enabled
+      | modules         |
+      | easy_breadcrumb |
```
imanoleguskiza: I don't understand why this module is required to create a view.

tests/features/taxonomy_browser.feature
```diff
@@ -0,0 +1,94 @@
...
+  @api
+  Scenario: Testing the block appears
+    When I am on the homepage
+    Then I should see the text "Tags"
+    Then I should see the text "economic"
+    Then I should see the text "sport"
+
+  @api
+  Scenario: Testing the filtering
+    When I am on the homepage
```
imanoleguskiza: The block will only work when used with a view. You should test it after navigating to the view you created before.

tests/features/taxonomy_browser.feature
```diff
@@ -0,0 +1,94 @@
...
+    And I should see the text "The configuration options have been saved."
+
+  @api
+  Scenario: Administrator user can add the block to a region
+    Given I visit "/admin/config/taxonomy_browser/settings_en"
+    And the checkbox "edit-taxonomy-browser-vocabulary-2" is checked
+    When I visit "/admin/structure/block_en"
+    Then I should see the text "taxonomy_browser: Tags"
+    When I go to "/admin/structure/block/manage/taxonomy_browser/taxonomy_browser_vocabulary_2/configure_en"
+    When I select "content" from "Ec_resp (default theme)"
```
imanoleguskiza: Please be careful that ec_resp is not the only theme we have in the platform. Since this is functional testing its ok to test it in ec_resp only but if that is the case you should add the @ec_resp_theme tag to the scenario

tests/features/taxonomy_browser.feature
VictorBonasa: I was trying to make some kind of "user stories" from the documentation (https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=MULTISITE&title=Taxonomy+browser+-+documentation) . Removed.

tests/features/taxonomy_browser.feature
```diff
@@ -0,0 +1,78 @@
...
+    Then I should see the text "* All changes are stored temporarily. Click Save to make your changes permanent. Click Cancel to discard your changes."
+    And I press the "Save" button
+    Then I should see the text "has been saved."
+    #Add the block to a region
+    Given I visit "/admin/config/taxonomy_browser/settings_en"
+    And the checkbox "edit-taxonomy-browser-vocabulary-2" is checked
+    When I visit "/admin/structure/block_en"
+    Then I should see the text "taxonomy_browser: Tags"
+    When I go to "/admin/structure/block/manage/taxonomy_browser/taxonomy_browser_vocabulary_2/configure_en"
+    When I select "Sidebar Left" from "Ec_resp (default theme)"
```
imanoleguskiza: You should add the tag @theme_wip to the scenario because it requires a region and this will cause issues with the ec_europa theme

VictorBonasa: Done! Thank you Imanol.

PR #1729, opened 2017-11-09
Nept 445: Improve behat test on contact feature

VictorBonasa:
### Description

Improve behat test on contact feature

### Change log

- Added: Added test for contact feature.

### Commands

No commands needed


Fefaine:
@VictorBonasa  The request is to test the feature 'contact_form' not the contributed module.
We assume contributed modules are tested already upfront
https://github.com/ec-europa/platform-dev/tree/master/profiles/common/modules/features/contact_form

VictorBonasa:
Thank you for the clarification @Fefaine . Contact feature fills in the name and email fields of the form with the same of the logged user, so "Test that existing email adress is filled in for authenticated users" and "Test that existing name is filled in for authenticated users" scenarios should be the same as for contact contrib module or if we assume contributed modules are tested already, this scenarios will not be necessary and can even be deleted.

I have rewrited "Test that email must be valid" and "Verify that mail is sent after the captcha is filled in when anonymous" to test the contact form provided by contact feature. However I cant test with captcha cause it breaks the purpose of Captcha which is specially designed to avoid automation and require human interaction. Any workaround or feedback on this would be great.

PR #1739, opened 2017-11-15
NEPT-1107: Add a patch that won't log missing content.

Fefaine:
### Description

When user decide to not show quicktabs missing content, missing content should not be logged.

### Change log

- Added: patch #2222805

### Commands

```sh
no command
```


aritomelo:
The PR is ready to be merged.

PR #1748, opened 2017-11-16
NEPT-1630: Reorder hook updates sequence in cce_basic_config.

Medi4tor:
### Description

After a merge ( https://github.com/ec-europa/platform-dev/pull/1714 ) the sequence of hook updates of cce_basic_config is broken.
we have :
219, 220 , 222, 222 , 223 , 224 
It has to be fixed asap.

### Change log
- Changed:
cce_basic_config.install


drupol:
👍

aritomelo:
OK, It's fine for me.

We should check this thinks while merging to avoid this kind of issues!

PR #1759, opened 2017-11-21
NEPT-1640: Feedback from QA, update file_entity to version + patch.

Fefaine:
### Description
Change the way we retrieve file_entity module, to be aligned with QA

### Change log

- Changed: multisite_drupal_standard.make file, changed to module version + patch instead of commit number


aritomelo:
Ready for the second review.

PR #1765, opened 2017-11-22
Nept 975 - Uninstall and remove multisite_dynamic_basetheme module

aritomelo:
### Description

As part of the upgrade to version 2.3 we noticed that in the past there was a procedure to keep several ec_resp versions and choose the base theme a site wanted to use.

### Change log
- Removed:
    The multisite_dynamic_basetheme module folder.

### Commands

```sh
drush dis multisite_dynamic_basetheme 
drush pm-uninstall multisite_dynamic_basetheme
drush cc all
```


gillesdeudon:
I close it; issue fixed in #1885

PR #1776, opened 2017-11-29
NEPT-1583: Upgrade Better Exposed Filters to 3.5.

gillesdeudon:
### Description

Upgrade Better Exposed Filters to 3.5.

### Change log

- Changed: The version of the contrib module Better Exposed Filters

### Commands

drush cc all


aritomelo:
Ready for CR+

PR #1782, opened 2017-11-29
NEPT-754: Fix line endings

jfhovinne:
### Description

Fix line endings

### Change log

- Fixed: Line endings

### Commands

No command


drupol:
Shouldn't we consider adding a .gitattributes file?

jfhovinne:
Indeed, created ticket to improve coverage as well.

jfhovinne:
@drupol BTW we have one, but to be checked.

PR #1786, opened 2017-11-30
NEPT-1453: Backport the Sanitize HTML enforcing in 2.2 at text format saving time

gillesdeudon:
### Description

Enforce the Sanitize HTML filter to any text format that is being saved in a site.

### Change log

- Added: Enforce the activation of the Sanitize HTML filters when a text format is saved.
- Changed: Add notification messages in the descriptions of inputs of the text format admin interface.

### Commands

```sh
drush cc all
```


Medi4tor:
Looks good.

PR #1787, opened 2017-11-30
NEPT-1677: Remove the check url feature from multisite wysiwyg and multisite drupal toolbox.

gillesdeudon:
### Description

Remove the "check url" feature from multisite wysiwyg and multisite drupal toolbox.

### Change log

- Removed: The "_multisite_drupal_toolbox_url_exists" callback
- Removed: The "multisite_wysiwyg_config.js" file

### Commands

```sh
drush cc all

```


Medi4tor:
Ok :+1:

PR #1799, opened 2017-12-12
NEPT-1704: Add patch to resolve message to throw errors on cron.

Fefaine:
### Description
Add patch to resolve message that throws errors on cron.

### Change log

- Added: Patch #2030101


Medi4tor:
It's OK for me.

PR #1803, opened 2017-12-12
NEPT-1701: [Poetry client] Sanitization of text fields on support-2.3

ademarco:
### Description
This PR will update Poetry client to version 0.3.5, which contains sanitization fix as per https://github.com/ec-europa/oe-poetry-client/pull/68

### Change log

- Added: -
- Changed: Update of Poetry client to 0.3.5
- Deprecated: -
- Removed: -
- Fixed: -
- Security: -

### Commands
Normally it would be enough to run composer install but I'm not sure about which one is the right command in this case since it's very deployment specific.

```sh
$ composer install
```

### DIFF
--- a/resources/composer.json
+++ b/resources/composer.json
@@ -3,7 +3,7 @@
   "prefer-stable": true,
   "require": {
     "guzzlehttp/guzzle": "^6.2.1",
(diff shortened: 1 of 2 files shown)


drupol:
👍

PR #1815, opened 2017-12-21
NEPT-366 Upgrade module scheduler to version 7.x-1.5

aritomelo:
### Description

Upgrade module scheduler to version 7.x-1.5
The update is quite big, so some test as a user should be done to check the impact, also some css/js around date field was changed.

### Change log

- Changed: Module version updated from 1.3 to 1.5

### Commands

```sh
drush updb

```


Medi4tor:
Ok

PR #1821, opened 2018-01-03
NEPT-272 Upgrade module L10n_update to version 7.x-2.2

aritomelo:
### Description
Upgrade module L10n_update from version 7.x-2.0 to 7.x-2.2

### Change log

- Changed: L10n_update module version.
- Removed: Included patch, already in the new version.

### Commands

```sh
drush updb
```


Medi4tor:
@aritomelo , Why did you also update webform module?

Medi4tor:
It seems like we don't need to do a 'drush updb' as there is no hook update... @aritomelo can you confirm and  update?

PR #1829, opened 2018-01-05
NEPT-1643: Update poetry plugin file to show correct info.

Medi4tor:
### Description

Update poetry plugin file to show correct info on detail of the job sent to translation.

### Change log

- Changed:
profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc

### DIFF
--- a/profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc
+++ b/profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc
@@ -303,8 +303,8 @@ class TMGMTPoetryTranslatorUIController extends TMGMTDefaultTranslatorUIControll
       '#default_value' => (!empty($job->settings['contacts']['responsable']) ? $job->settings['contacts']['responsable'] : $contacts['responsable']),
     );
 
-    // Load translator settings : Organization.
-    $organization = $translator->getSetting('organization');
+    // Load translator settings from submitted job: Organization.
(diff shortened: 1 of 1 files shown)


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/plugins/tmgmt_poetry.ui.inc
```diff
@@ -303,8 +303,8 @@ class TMGMTPoetryTranslatorUIController extends TMGMTDefaultTranslatorUIControll
       '#default_value' => (!empty($job->settings['contacts']['responsable']) ? $job->settings['contacts']['responsable'] : $contacts['responsable']),
     );
 
-    // Load translator settings : Organization.
-    $organization = $translator->getSetting('organization');
+    // Load translator settings from submitted job: Organization.
+    $organization = ($job->settings['organization']) ? $job->settings['organization'] : $translator->getSetting('organization');
```
aritomelo: Here I would suggest we check whether or not the array key "organization" exists as from tmgmt the setting is an array and the key is not set by default, so nothing guarantee that it always exists.

Medi4tor: Hi @aritomelo I updated  the code in order to avoid the 'undefined index" on the array.
Please check, ammend and validatre. 
K.R.

aritomelo:
Ready for CR +1!

I was able to check the issue and afterwards the solution!

PR #1839, opened 2018-01-10
NEPT-1325: Add patch for rules_init() and cache rebuilding are broken.

Fefaine:
### Description
Add patch for rules_init() and cache rebuilding are broken.

### Change log

- Added: Patch for issue #2851567


Medi4tor:
Ok for me.

PR #1840, opened 2018-01-10
NEPT-1325: Upgrade rules module to version 2.10

Fefaine:
### Description
Upgrade rules module to version 2.10

### Change log

- Changed: Version of rules module.

### Commands

```sh
drush updb

```


Medi4tor:
Ok for me.

PR #1842, opened 2018-01-11
NEPT-1750: Add a patch that fix the entityreference language regression.

Fefaine:
### Description

Add a patch that fix the entityreference language regression.

### Change log

- Added: Patch 2850416


aritomelo:
Ready for +1

PR #1845, opened 2018-01-12
NEPT-1753: Concatatenate the LANGUAGE_NONE instead of put between doubble quotes

gillesdeudon:
### Description

Fix issue with the misuse of LANGUAGE_NONE in the multisite_drupal_language_negociation_views_query_alter

### Change log

- Fixed: misuse of LANGUAGE_NONE in the multisite_drupal_language_negociation_views_query_alter


Medi4tor:
OK !

PR #1849, opened 2018-01-16
NEPT-1722: Add a patch to fix the condition current vs previous state

Fefaine:
### Description
Add a patch to fix the condition current vs previous state

### Change log

- Added: Patch to fix issues on previous state condition


jfhovinne:
Please do not forget documenting the patch in the makefile.
Regarding commit message, no dot (.) at the end of the sentence.
Same regarding PR title. Thanks !

PR #1850, opened 2018-01-16
NEPT-1761: Fix building issues on release-2.5

gillesdeudon:
### Description

Fix issues that blocks the build in release-2.5

### Change log

- Changed: Fix CS in multisite_drupal_toolbox.install
- Removed: The useless patch: rendered_entity_is_not_language_aware_again-2850416-3.patch

### Commands
No command needed.


Medi4tor:
PR is green! OK for me

gillesdeudon:
I will make a simple merge instead of a squash and merge because the 2 commits are on different topics; even if they concern the same ticket.

PR #1852, opened 2018-01-16
NEPT-1666: Create a new cart page

dxvargas:
### Description

Create a new cart page that will serve as dashboard.

### Change log

- Added: New page

### Commands

```sh
no commands
```


voidtek:
next step is here https://github.com/ec-europa/platform-dev/pull/1854

PR #1866, opened 2018-01-25
NEPT-172: Improve URL rewrite suffix function by adding some caching techniques.

Fefaine:
### Description
Improve URL rewrite suffix function by adding some caching techniques.

### Change log

- Changed: Refactored nexteuropa_multilingual_language_negotiation_url_rewrite_callback

### Command

drush cc all


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc
```diff
@@ -9,51 +9,53 @@
  * Callback functions for the nexteuropa language negotiation providers.
  */
 function nexteuropa_multilingual_language_negotiation_provider_callback($languages) {
-  $language_url = FALSE;
   $current_path = request_path();
 
+  // $_GET['q'] might not be available at this time, because
```
dxvargas: $_GET['q'] can also be empty for the domain name of the site. The code is correct for that case also, but the comment seems to ignore it, which is confusing.

gillesdeudon:
I close the PR as it is obsolete.

PR #1871, opened 2018-01-30
NEPT-1189: change block name in migrate module and nexteuropa_communities.

voidtek:
### Description

Theme the member list inside the community feature

### Change log

- Added: Theme the member list inside the community feature

### Commands

```sh
no commands

```


profiles/common/modules/features/communities/communities.views_default.inc
```diff
@@ -547,7 +547,7 @@ function communities_views_default_views() {
   $handler->display->display_options['pager']['type'] = 'some';
   $handler->display->display_options['pager']['options']['items_per_page'] = '5';
   $handler->display->display_options['pager']['options']['offset'] = '0';
-  $handler->display->display_options['style_plugin'] = 'list';
+  $handler->display->display_options['style_plugin'] = 'ec_europa_link_blocks';
```
voidtek: This style is only available if the module nexteuropa_formatters_views is enabled, but this module is not enable by default.
For ec_resp, we need to keep the style "list" and for ec_europa, we can put this style.

profiles/common/modules/features/communities/communities.views_default.inc
```diff
@@ -749,7 +749,7 @@ function communities_views_default_views() {
   $handler->display->display_options['exposed_form']['type'] = 'basic';
   $handler->display->display_options['pager']['type'] = 'some';
   $handler->display->display_options['pager']['options']['items_per_page'] = '5';
-  $handler->display->display_options['style_plugin'] = 'default';
+  $handler->display->display_options['style_plugin'] = 'ec_europa_link_blocks';
```
voidtek: This style is only available if the module nexteuropa_formatters_views is enabled, but this module is not enable by default.
For ec_resp, we need to keep the style "list" and for ec_europa, we can put this style.

AlbaParrilla:
It has been modified the function _nexteuropa_ecl_migrate_enable_ec_europa_blocks() because it gave a PDO Exception when trying to change an old site from ec_resp to ec_europa by using the following drush command: 
drush php-eval '_nexteuropa_ecl_migrate_blocks();'

PR #1874, opened 2018-02-01
NEPT-1727: Upgrade Taxonomy Term Reference Tree Widget to 1.11.

gillesdeudon:
### Description

Upgrade the "Taxonomy Term Reference Tree Widget" module to 1.11.

### Change log

- Changed: the make file to point to the version 1.11 of "Taxonomy Term Reference Tree Widget" 

### Commands

drush cc all


gillesdeudon:
I close the PR because it does not point to the right ticket

PR #1878, opened 2018-02-01
NEPT-1909: Update versions of atomium (2.6) and ec_europa (0.0.8).

gillesdeudon:
### Description

Update versions of atomium (2.6) and ec_europa (0.0.8).

### Change log

- Changed: make file and build.properties.dist

### Commands

drush cc all

### DIFF
--- a/build.properties.dist
+++ b/build.properties.dist
@@ -14,7 +14,7 @@ platform.site.theme_default = ec_resp
 ecl.version = 0.18.0
 
 # The default EC Europa theme release
-ec_europa.version = 0.0.6
+ec_europa.version = 0.0.8
 
 # The default Atomium and Europa theme build properties.
 # Used only if default theme is set to "ec_europa".
--- a/resources/multisite_drupal_standard.make
+++ b/resources/multisite_drupal_standard.make
@@ -1016,9 +1016,9 @@ projects[ec_resp][download][tag] = 2.3.4
 projects[atomium][type] = theme
 projects[atomium][download][type] = git
 projects[atomium][download][url] = https://github.com/ec-europa/atomium.git
-projects[atomium][download][branch] = 7.x-2.x
+projects[atomium][download][tag] = 7.x-2.6
 
 projects[ec_europa][type] = theme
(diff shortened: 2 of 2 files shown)


build.properties.dist
```diff
@@ -14,7 +14,7 @@ platform.site.theme_default = ec_resp
 ecl.version = 0.18.0
```
voidtek: We need to update the ecl version, according the documentation of ec_europa.

resources/multisite_drupal_standard.make
```diff
@@ -1016,9 +1016,9 @@ projects[ec_resp][download][tag] = 2.3.4
 projects[atomium][type] = theme
 projects[atomium][download][type] = git
```
voidtek: Atomium is in d.org as a contrib module.

resources/multisite_drupal_standard.make
```diff
@@ -1016,9 +1016,9 @@ projects[ec_resp][download][tag] = 2.3.4
...
 projects[atomium][download][url] = https://github.com/ec-europa/atomium.git
-projects[atomium][download][branch] = 7.x-2.x
```
voidtek: According with the documentation - https://webgate.ec.europa.eu/fpfis/wikis/display/MULTISITE/Contributed+modules#version, it's not necessary to change this version.

build.properties.dist
gillesdeudon: I make the modification.

resources/multisite_drupal_standard.make
gillesdeudon: Just for the record:
I just adapted the part requested by the ticket. 
The rest was already like that before and I did not change because we do not have rules for this kind of case.

I will adapt it.

resources/multisite_drupal_standard.make
jonhy81: HI @voidtek, look deeper in the example you are missing the commit hash.

gillesdeudon:
I renamed the branch, the new PR is #1879

PR #1880, opened 2018-02-02
NEPT-1781: Show/hide Request Translation button

dxvargas:
### Description

Only one button should be visible, Sent to Card or Request Translation.

### Change log

- Changed: hook_form_alters for translation form

### Commands

```sh
no command needed
```


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/tmgmt_dgt_connector.inc
```diff
@@ -244,3 +244,24 @@ function _tmgmt_dgt_connector_translator_access($op, array $translator, array $a
...
+ *
+ * @param object $entity
+ *   The entity to check.
+ * @param array $cart
+ *   The cart to be generated.
+ *
+ * @return bool
+ *   Returns true if 'Send to cart' is possible, false otherwise.
+ */
+function _tmgmt_dgt_connector_translation_possible($entity, array $cart) {
```
imanoleguskiza: The name of the function is a bit confusing it doesn't check if translation is possible, but rather if sending to cart is possible.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/tmgmt_dgt_connector.module
```diff
@@ -85,19 +85,14 @@ function tmgmt_dgt_connector_nexteuropa_poetry_notification_status_updated($mess
...
-  $cart = $form_state['tmgmt_cart'];
-  list($entity_id) = entity_extract_ids($entity->entity_type, $entity);
-
-  $source_data = _tmgmt_dgt_connector_get_source_data($cart['plugin'], $entity->entity_type, $entity_id);
-  $length = _tmgmt_dgt_connector_count_source_data($source_data);
+  // Set plugin for #cart .
+  $form_state['tmgmt_cart']['plugin'] = 'workbench_moderation';
 
-  if ($length > TMGMT_DGT_CONNECTOR_MAX_SMALLJOB_LENGTH) {
+  if (!_tmgmt_dgt_connector_translation_possible($form_state['entity'], $form_state['tmgmt_cart'])) {
```
imanoleguskiza: This whole interaction is a bit weird, we remove the default cart button if the length is small enough, but keep it if not? For me we either keep the old cart or we don't, but both at the same time seems confusing and unnecesary.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_poetry/tmgmt_poetry.module
```diff
@@ -863,6 +856,39 @@ function tmgmt_poetry_form_tmgmt_entity_ui_translate_form_alter(&$form, &$form_s
...
+  // Checking if we have a node with the appropriate moderation state.
+  // Getting the moderation states which allows to request the translation.
+  $states = array_values(_tmgmt_workbench_available_states());
+  $state = $form_state['entity']->workbench_moderation['current']->state;
+
+  if (in_array($state, $states, 'strict')) {
+    $translations_possible = array('regular');
+  }
+
+  // If Cart translation is possible, Regular translation shouldn't be possible.
```
imanoleguskiza: Here we assume the existence of a cart module and act accordingly, but that is not the case. Part of this functionality should go into the cart module.
It should be the cart modules responsibility to show or hide its button.

dxvargas: It's actually what's happening. Cart module will add it's own button.

But in the situation when cart module adds it's own button, we don't want to show the regular poetry translation button. So Poetry module needs to check that here.

PR #1885, opened 2018-02-07
   NEPT-975: Uninstall and remove multisite_dynamic_basetheme module

gillesdeudon:
### Description

As part of the upgrade to version 2.3 we noticed that in the past there was a procedure to keep several ec_resp versions and choose the base theme a site wanted to use.

### Change log
- Removed:
    The multisite_dynamic_basetheme module folder.


Medi4tor:
Looks OK for me. The module is removed.

phpcs-ruleset.xml
```diff
@@ -37,6 +37,9 @@
     <!-- This module is not supported by Core team so we exclude it from QA -->
     <exclude-pattern>profiles/*/modules/custom/nexteuropa_newsroom/</exclude-pattern>
 
+    <!-- This module is not supported by Core team so we exclude it from QA, to avoid phpcs issue -->
+    <exclude-pattern>profiles/*/modules/custom/nexteuropa_newsroom/</exclude-pattern>
+
```
drupol: I see two issues here:

1. this is out of scope
2. it's duplicated

Maybe a wrong merge ?

Medi4tor: Nice catch Pol.
We will have to remove that.

PR #1892, opened 2018-02-09
NEPT-1804: Add available translations in variable theme.

Medi4tor:
### Description

Add available translations in variable theme.

### Change log

- Changed:
     profiles/common/modules/custom/language_selector_site/language_selector_site.module


Medi4tor:
Change made in  .make file.

PR #1920, opened 2018-02-28
NEPT-1671: Add patch for simplenews statistics.

Medi4tor:
### Description

Add patch for simplenews statistics.
To prevent simplenews module throws an SQL error when creating new newsletters

### Change log
- Changed:
resources/multisite_drupal_standard.make


gillesdeudon:
I close because the fix is in PR #1724.

PR #1942, opened 2018-03-19
NEPT-1879: Improve the control on the pop-up display limitations.

gillesdeudon:
### Description

Limit the use of HTML tags to inline one in the nexteuropa_geofield fields.

### Change log

- Changed: Add filter to limit the HTML tags uses in nexteuropa_geofield fields.

### Commands

drush cc all


AlbaParrilla:
approve code review

PR #1963, opened 2018-04-04
NEPT-582: Add test for multisite_create_button functionality for both profiles

enriquelacoma:
### Description

Create behat test for creation button and selector

### Change log

- Added: file tests/features/multisite_create_button.feature and steps definitions to tests/src/Context/DrupalContext.php


tests/src/Context/DrupalContext.php
```diff
@@ -347,5 +347,43 @@ public function getElementsByAttribute(Element $source_element, $tag, $attribute
...
   }
+  /**
+   * Click on a selector element.
+   *
+   * @param string $arg1
+   *   Selector css.
+   *
+   * @Then I click on selector :arg1
+   */
+  public function iClickOnSelector($arg1) {
```
imanoleguskiza: I think this function should be labeled "iClickOnElement" because it could potentially be used to filter based on any css. Also the step should change accordingly ("@Then I click on the element :arg1")

resources/drupal-core.make
```diff
@@ -2,7 +2,7 @@ api = 2
 core = 7.x
 
 projects[drupal][type] = "core"
-projects[drupal][version] = "7.57"
+projects[drupal][version] = "7.58"
```
imanoleguskiza: Is this update required? It doesn't seem like it belongs in this PR

imanoleguskiza (dismissed):

aritomelo (approved):
All good.

ku-enza (approved):
All green.

PR #1970, opened 2018-04-10
NEPT-1259: mpdf library: reduce its weight

enriquelacoma:
### Description

Reduce the size of the mpdf library by using the fonts from the 2.2 release

### Change log

- Added: Added to the build process the fonts from https://github.com/mpdf/mpdf/archive/v5.7.4.zip

### Commands

Requires building the platform


enriquelacoma:
Won't be inluded on 2.5, to be added on 2.6 (https://github.com/ec-europa/platform-dev/pull/1973)

PR #2014, opened 2018-04-21
Assign CEM role to ECAS user of a defined group

gillesdeudon:
### Description

Assign automatically the CEM role to ECAS user of a defined group.
Add more permissions to the cem role
Align the cam permissions on contributor ones in the feature definitions.

### Change log

- Added: hook_ecas_sync_user_info() in the ecas module with documentation.
- Changed: The default cem role permissions and move the code to cce_basic_config module.

### Commands

None

### DIFF
--- a/profiles/common/modules/custom/ecas/ecas.api.php
+++ b/profiles/common/modules/custom/ecas/ecas.api.php
@@ -1,0 +1,63 @@
+<?php
+
+/**
+ * @file
+ * Documentation file for the module API.
+ */
+
+/**
+ * Act on the list of user data value to save of the Ecas user.
+ *
+ * @param array $edit
+ *   The list array of user data values. it is the same "$edit" parameter as
+ *   in user_save().
+ * @param object $user
+ *   A user object.
+ * @param array $user_info
+ *   An associative array with the following interesting keys:
+ *   - mail: mail address.
+ *   - givenname: first name.
+ *   - sn: last name.
+ *   These values will be used to fill fields/profiles/...
+ * @param array $args
+ *   Extra parameters, not used directly in this function but passed to the
+ *   info_ecas_update() hook.
+ */
+function hook_ecas_sync_user_info(array &$edit, $user, array $user_info, array $args) {
+  $role = user_role_load_by_name('role_name');
+  $user_roles = (isset($user->roles)) ? $user->roles : array();
+  $user_roles[$role->rid] = $role->name;
+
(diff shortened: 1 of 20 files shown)


profiles/common/modules/custom/ecas/ecas.api.php
```diff
@@ -0,0 +1,63 @@
...
+ *   An associative array with the following interesting keys:
+ *   - mail: mail address.
+ *   - givenname: first name.
+ *   - sn: last name.
+ *   These values will be used to fill fields/profiles/...
+ * @param array $args
+ *   Extra parameters, not used directly in this function but passed to the
+ *   info_ecas_update() hook.
+ */
+function ecas_group_sync_info_ecas_update($user, array $user_info, array $args) {
```
imanoleguskiza: I think here you mean "hook_info_ecas_update"?

profiles/common/modules/custom/ecas/ecas.api.php
```diff
@@ -0,0 +1,63 @@
...
+ *   An associative array with the following interesting keys:
+ *   - mail: mail address.
+ *   - givenname: first name.
+ *   - sn: last name.
+ *   These values will be used to fill fields/profiles/...
+ * @param array $args
+ *   Extra parameters, not used directly in this function but passed to the
+ *   info_ecas_update() hook.
+ */
+function hook_ecas_sync_user_info(array &$edit, $user, array $user_info, array $args) {
```
imanoleguskiza: Should we not keep the same argument ordering through functions? ($user, $user_info, $edit, ..... as we see in functions like ecas_sync_profile_core and even in user_save)

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -351,23 +349,66 @@ function ecas_sync_profile_core(&$user, array $user_info) {
...
 }
 
+/**
+ * Implements hook_ecas_sync_user_info().
+ *
+ * Assign the "CEM" role assigned to the user if they belong to the COMM_CEM.
+ */
+function ecas_ecas_sync_user_info($user, array $user_info, array &$edit, array $args) {
+  $cem_group = 'COMM_CEM';
+  $cem_role_name = 'cem';
```
dxvargas: 'COMM_CEM' and 'cem' strings should be constants.
We already have a role name defined in a constant: NEXTEUROPA_EDITORIAL_TEAM_MEMBER_ROLE.

PR #2019, opened 2018-04-26
NEPT-1932: Update Drupal core to 7.59

Medi4tor:
### Description

Update Drupal core to 7.59

### Change log

- Security: Update Drupal core to 7.59

### Commands

No command.


drupol:
I think there is an issue in your pull request mate.

PR #2030, opened 2018-05-02
Nept-1882

imanoleguskiza:
### Description

Remove the default TMGMT Cart

### Change log

- Removed: Remove the default TMGMT Cart


profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/README.md
```diff
@@ -1,71 +1,70 @@
...
 
-Then, in the ```Small Jobs Cart (admin/tmgmt/dgt_cart)```, you can send a bundle to be translated.
-Each bundle is a set of items having the same target languages set.
-Before sending a bundle for translation, please edit the bundle, then revise and complete
-the context information: Context URL should be filled and also any other
-hints that may help the person in charge of translation to understand the
-context where the item will be displayed.
+# Usage
 
+Requesting translations for content whose length is bigger than 300 follows the usual workflow
```
dxvargas: Confusing and "whose" is used more when referring to a person. Suggestion:
"Requesting translations for content with length bigger than 300 characters"

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/README.md
```diff
@@ -1,71 +1,70 @@
...
+of the TMGMT module. Please refer to the documentation of the module for more details.
+
+In order to bundle items together a user must use the custom DGT Cart which replaces the
```
dxvargas: Using the "custom DGT Cart" is the recommended way of proceeding here. But in fact the TMGMT cart is also a possibility. So I would put here "a user should use the custom DGT Cart". It's just my opinion, if you don't agree please ignore.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/README.md
```diff
@@ -1,71 +1,70 @@
...
+string type.
+After selecting the desired strings, they can be sent to the cart itself by selecting the target
+languages and clicking the "Send to cart" button.
+
+#### DGT Cart
+
+The custom DGT cart can be accessed through the following url "admin/tmgmt/dgt_cart".
+The cart shows all the currently created bundles.
+Each bundle is created based on the target languages of the items sent into the cart.
+Once in a bundle, the items inside them can be edited and extra metadata added by using
```
dxvargas: "them" is wrong because it is a single bundle.
I suggest:
"The Edit option allows to see the items inside a bundle, edit its context data and also to remove items.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/README.md
```diff
@@ -1,71 +1,70 @@
...
+edit option.
+The Discard option will delete the bundle and remove all the items in it.
+The Send option will create a Translation request which will then follow the usual workflow
+of the TMGMT module.
 In the translation checkout page, the translator ```TMGMT DGT Translator``` must be used.
```
dxvargas: This last sentence should not be under "DGT Cart". It is related to the Checkout page.

profiles/common/modules/features/nexteuropa_dgt_connector/tmgmt_dgt_connector/README.md
imanoleguskiza: I think in this case I will not change it simply to imply that the DGT cart is the way to go from now on (even though as you say they could potentially not use it)

PR #2047, opened 2018-05-17
NEPT-1950: Add registry_autoload patch of the 2870868 issue

gillesdeudon:
### Description

Fixing issue with the class loading during the registry rebuild.

### Change log

- Changed: Add registry_autoload 2870868 patch to  multisite_drupal_standard.make

### Commands

drush cc all


verbruggenalex:
This looks fine. 👍  If it solves the issue with entity classes. The only patch I was not really in favor for was:
https://www.drupal.org/files/issues/entitycache_fix_upgrade_path-2441965-62.patch

CC @gillesdeudon

PR #2073, opened 2018-06-01
NEPT-520: Prevent EU Login to create 2 users with the same email.

ku-enza:
### Description

Prevent EU Login to create 2 users with the same email

### Change log

- Changed: profiles/common/modules/custom/ecas/includes/ecas.inc

### Commands

drush cc all


profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -673,13 +710,32 @@ function _ecas_force_login() {
...
 
-  $output .= '<h2 class="ecl-heading ecl-heading--h2">' . t('Account blocked') . '</h2>';
-  $output .= '<p>' . variable_get('ecas_warning_message', t('Your account is not yet approved')) . '</p>';
+  switch ($reason) {
+    case 'blocked':
+      $output .= '<h2 class="ecl-heading ecl-heading--h2">' . t('Account blocked') . '</h2>';
+      $output .= '<p>' . variable_get('ecas_warning_message', t('Your account is not yet approved')) . '</p>';
+      break;
 
+    case 'notfound':
```
dxvargas: This error is shown when the email exists for another username. Both $reason value ('notfound') and message ('Sorry, we cannot found your account.') here are misleading. I don't understand, is it on propose?!

If it is to remain, just fix the grammar:
'Sorry, we cannot found your account.' should be 'Sorry, we could not find your account.'

But I suggest to really explain what happened, like this the user knows that it should login using the initial EU Login authentication method.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -374,128 +375,144 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
...
+    }
 
-  // If we don't have a user register them.
-  if (!$account && !empty($ecas_name)) {
+    if (phpCAS::isAuthenticated() && (($force_login === ECAS_GATEWAY) || ($force_login === ECAS_LOGIN))) {
+      $ecas_name = phpCAS::getUser();
+      // @codingStandardsIgnoreStart: NEPT-1332: To be fixed during refactoring.
+      $ecas_attributes = phpCAS::getAttributes();
+      $ecas_groups = $ecas_attributes['cas:groups'];
+      $ecas_email = $ecas_attributes['cas:email'];
```
dxvargas: Here it's better to already use drupal_strtolower(trim()). Like that we have $ecas_email ready to consume whenever we need.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -155,7 +156,7 @@ function _ecas_menu(array &$items) {
  * Apply ECAS logout logic when logging out because of autologout rules.
  */
 function ecas_user_logout($account) {
-  if (isset($_GET['q']) && $_GET['q'] == 'autologout_ahah_logout/alt') {
+  if (isset($_GET['q']) && $_GET['q'] == 'autologout_ahah_logout') {
```
dxvargas: I don't think this change was intended. And it's not correct. Please remove it or justify.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -374,128 +375,144 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
 function ecas_login_check() {
   global $user;
 
+  $force_login = _ecas_force_login();
   if ($user->uid) {
     // The user has an uid, and it is > 0 (which is anonymous user's uid) do
     // nothing because user is already logged into Drupal.
     return;
   }
+  elseif ($force_login != ECAS_NO_LOGIN) {
```
dxvargas: Please don't do this. Before it was much better.
Like this, all the code inside needs to be indented. And it's a long elseif, you have to scroll to see where it ends and if there's another else in the end (it doesn't).
Also we don't need a elseif when if makes method to return.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -374,128 +375,144 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
...
+      // These variables will be checked by later conditions.
+      $account = FALSE;
+      $ecas_name = NULL;
+      $ecas_email = NULL;
+    }
 
-    $account = ecas_create_user($ecas_name);
+    // Check that the email is not present for another account.
+    if (!$account && !empty($ecas_email)) {
+      $account_check = user_load_by_mail(drupal_strtolower(trim($ecas_email)));
```
dxvargas: I suggest drupal_strtolower(trim()) to be applyed once when $ecas_email is given its value. Then every time it's needed we know it's correct.
It's only a suggestion because we only use $ecas_email once , so it's more a best practice with no real impact (ATM).

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -673,13 +710,32 @@ function _ecas_force_login() {
 
 /**
  * Page callback; displays the 'ecas_warning_page' page.
+ *
+ * @param string $reason
+ *   REason for warning message.
```
dxvargas: Minor typo: E should be e.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -562,6 +596,9 @@ function ecas_block_view($delta = '') {
       else {
         $output = '<div class="user-login-block">' . "\n";
 
+        // NOTE: special care needs to be taken because on pages with forms,
+        // such as node and comment submission pages, the $edit variable might
+        // already be set.
```
dxvargas: I don't understand this comment.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -554,6 +585,9 @@ function ecas_block_view($delta = '') {
       if (!$user->uid) {
         $output = '<div class="user-login-block">' . "\n";
 
+        // NOTE: special care needs to be taken because on pages with forms,
+        // such as node and comment submission pages, the $edit variable might
+        // already be set.
```
dxvargas: I don't understand this comment.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -374,128 +375,144 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
...
-    // Case when account is blocked.
-    if ($account->status == 0) {
-      _ecas_session_end();
-      drupal_goto('ecas_warning_page');
+      // Also set an extra parameter for ecas_sync_user_info() -- this allows
+      // other modules to detect a user creation.
+      $ecas_sync_user_info_args['ecas_user_creation'] = TRUE;
     }
-    // When returning with same ECAS login as before closing Drupal session.
-    else {
```
dxvargas, line 531: Why did you delete this logic?! It's needed here.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -374,128 +375,144 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
...
+      if ($account->status == 0) {
+        // Case when account is blocked.
+        watchdog('user', 'Login denied for %ecas_name as blocked user.',
+          array('%ecas_name' => $ecas_name));
+        session_destroy();
 
-        $url = phpCAS::getServerLogoutURL() . '?url=' . _ecas_get_url_site_frontpage();
-        $msg = t('Your EU Login session needs to be renewed. For security reasons, please <a href="@url">logout and login again</a>.', ['@url' => $url]);
-        drupal_set_message($msg, 'error');
+        $user = drupal_anonymous_user();
```
dxvargas: Why here we need to set $user = drupal_anonymous_user(); ?

In other places where login fails, we have only this methods:
watchdog()
session_destroy()
drupal_goto()

It should be the same everywhere, or explain why here it's different please.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -514,13 +531,27 @@ function ecas_login_check() {
...
+}
 
+/**
+ * Internal function to unset destination from request.
+ */
+function _ecas_unset_destination() {
+  // We cannot ensure that the user is the same as in Ecas,
+  // So we absolutely want to redirect the user
+  // to the ECAS warning page - thus, we unset any previously set destination.
+  if (isset($_REQUEST['destination'])) {
```
dxvargas: Function drupal_goto() works with $_GET['destination'] .
I'm not sure if changing $_REQUEST['destination'] you have the desired effect.

dxvargas (changes requested):

enriquelacoma (approved):

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

PR #2077, opened 2018-06-06
EMA-513: Fix missing warning about missing theme.

drupol:


tatarbj:
Even i know my thumbs up doesn't count for the internal workflow of the team, still i just would like to let the members know, i've tested it on a clean ema instance and it correctly resolves the original reported issue!
Good job!
👍

PR #2078, opened 2018-06-06
Fix warning about missing theme.

richardcanoe:


drupol:
See #2080 and #2081 .

PR #2086, opened 2018-06-11
NEPT-820: Improve piwik module.

aritomelo:
### Description

Piwik site path is not taken into account for rendering the analytic script.

### Change log

- Added:
 Check whether or not the paths set by site administrator match the current URL so the analytic script will be rendered.


gillesdeudon (dismissed):
Ok for me

dxvargas (dismissed):

enriquelacoma (dismissed):

Fefaine:
This change if frozen for now

PR #2087, opened 2018-06-11
NEPT-820: Improve Piwik custom module

aritomelo:
NEPT-820
Description
Piwik site path is not taken into account for rendering the analytic script.

Change log
Added:
Check whether or not the paths set by site administrator match the current URL so the analytic script will be rendered.
Improve piwik custom module, only load code when a domain is matched, return 404 & 403 also when search is not enabled.

Behat Test update the way is taken in consideration the paths check.


profiles/common/modules/custom/nexteuropa_piwik/nexteuropa_piwik.module
```diff
@@ -264,20 +264,38 @@ function _nexteuropa_piwik_visibility_pages() {
 
     $visibility = variable_get('nexteuropa_piwik_visibility_pages', 0);
     $setting_pages = variable_get('nexteuropa_piwik_pages', NEXTEUROPA_PIWIK_PAGES);
+    $site_path_array = _nexteuropa_piwik_site_path_array();
+    $current_url = $_SERVER['HTTP_HOST'] . request_uri();
+    $matching_root = FALSE;
+
+    foreach ($site_path_array as $path) {
+      // Is the path match hte domain?
+      if (strpos($current_url, trim($path)) !== FALSE) {
```
dxvargas: Actually trim() doesn't make any difference here.

gillesdeudon:
Due to the ticket is not a clone and the ticket is major, this PR is closed and the fix is only ported from 2.5, see PR #2086

PR #2094, opened 2018-06-18
NEPT-1991: Update themes EC Europa and Atomium

jfhovinne:
### Description

Minor update of the themes.

### Change log

- Changed: Themes EC Europa and Atomium

### Commands

drush cc all


jfhovinne:
+1

PR #2095, opened 2018-06-18
NEPT-1987: Fix in i18n Fatal Error Call to undefined method i18n_object_wrapper::strings_update()

gillesdeudon:
### Description

Add patch of D.o issue 2082573 related to I18n module ini order to fix the fatal error "call to undefined method".
This fatal error occurs on some sub-sites when they install/enable a new module on their site.
The patch fixes the issue by resetting the static caches used by I18n module. 

### Change log

- Changed: multisite_drupal_standard.make, add the line declaring the 2082573 patch

### Commands

No drush command


jfhovinne:
I would suggest a more readable PR title and description, better describing the PR purpose (what it changes, fixes, ...).

PR #2127, opened 2018-07-25
NEPT-1606: Implement drone for php5.6

voidtek:
### Description

Update pipeline to test PHP 7 compatibility

### Change log

- Added: Drone and Docker-Composer file.

### Commands

no commands


aritomelo:
I have enabled this branch because the Atomium team wants to test on the platform and this will help us to speed up the process.

PR #2141, opened 2018-08-08
NEPT-2021: Block EU login if the e-mail is not present in the profile

gillesdeudon:
### Description

Implement a control on the presence of the e-mail address in the EU login user profile.
If it is not there, the Drupal user creation is stopped and the user is redirected to an error page with information about how to unblock the situation.

### Change log

- Added:
  - profiles/common/modules/custom/ecas/includes/ecas.inc: Add control on the presence of the e-mail and the redirection process
  - profiles/common/modules/custom/ecas/ecas.variable.inc: Declaring the translatable variables
  - profiles/common/modules/custom/ecas/includes/ecas.admin.inc: Add 2 field in the admin form for the error messages to display; one for a EU Login issue and one when the issue is related to social network.
  - profiles/common/modules/custom/ecas/ecas.install and /ecas.install.inc: Set the default value for the 2 new variable at install time.
  - profiles/common/modules/custom/ecas/ecas.tokens.inc: Declare new ecas specific tokens, 5 allowing  to introduce default illustration images in the error page and 2 to include URLs linked to the EU Login interface.
- Changed:
  - /profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc: Modify the language suffix rewrite callback to include the ecas warning and error pages in the suffix treatment and keep the other ecas page oustide.

### Commands

drush cc all

At site level, they must be notified that the ecas variable are translatable now if these variables are flagged as translatable in i18n_variable module.


profiles/common/modules/custom/ecas/includes/ecas.constants.inc
```diff
@@ -49,3 +49,28 @@ define('ECAS_GATEWAY', 2);
 define('ECAS_DEFAULT_PAGES', '');
 define('ECAS_LOGIN_EXLUDE_PAGES', 0);
 define('ECAS_LOGIN_INCLUDE_PAGES', 1);
+
+/**
+ * ECAS group for CEM team users.
+ */
+define('ECAS_CEM_ECAS_GROUP', 'COMM_CEM');
```
dxvargas: Why this here now?!

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -180,10 +184,20 @@ function ecas_menu_check() {
...
+  drupal_goto($account_request_url);
+}
+
+/**
+ * Gets the URL for the EU Login account creation form.
+ *
+ * @return string
+ *   The creation form URL.
+ */
+function ecas_get_account_request_url() {
```
dxvargas: It's an internal function, it should be prefixed with "_".

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -354,7 +369,29 @@ function ecas_sync_user_info(&$user, array $user_info, array $args) {
   // Let the opportunity to modules to alter user roles.
   drupal_alter('ecas_user_roles', $edit['roles'], $user, $user_info, $args);
 
+  // We just gave the possibility to others modules to treat the case, now we
```
dxvargas: Function `ecas_sync_user_info` is also used in `ecas_import_users_show_user_content_form_submit` and in that case we don't want to redirect the user.

If a login comes without an associated email we should reject the login and don't sync any info, so we shouldn't even get here in this code.

profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -826,3 +914,99 @@ function ecas_get_user_department($user_info) {
...
+/**
+ * Gets the error message value for the specified variable.
+ *
+ * @param string $variable_name
+ *   The name of the variable related to the error message.
+ *
+ * @return array|bool
+ *   The error message value; or FALSE if not found.
+ */
+function _ecas_get_error_message_value($variable_name) {
```
dxvargas: I don't agree with this approach but since it's not a technical opinion (it's an opinion about the overall strategy that affects the all solution), I will give my opinion in Jira ticket.

profiles/common/modules/custom/ecas/includes/ecas.constants.inc
```diff
@@ -49,3 +49,28 @@ define('ECAS_GATEWAY', 2);
...
+
+/**
+ * Name of the login message variable.
+ */
+define('ECAS_LOGIN_MESSAGE', 'ecas_login_message');
+
+/**
+ * Name of the warning message variable.
+ */
+define('ECAS_WARNING_MESSAGE', 'ecas_warning_message');
```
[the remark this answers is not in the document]
gillesdeudon: 👍

profiles/common/modules/custom/ecas/includes/ecas.constants.inc
gillesdeudon: I grouped the constants that I have under the hand there. Apparently the "ecas.constants.inc" is designed for this.

profiles/common/modules/custom/ecas/includes/ecas.inc
gillesdeudon: I was not planning to make it private because I do not see why it could not be part of the module API.
Is this problematic?

profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc
```diff
@@ -41,8 +41,9 @@ function nexteuropa_multilingual_language_negotiation_url_rewrite_callback(&$pat
     return;
   }
 
-  // Don't add the language suffix for sitemap.*** and ecas.
-  if (strpos($path, 'sitemap.') === 0 || strpos($path, 'ecas') === 0) {
+  // Don't add the language suffix for "sitemap.***" and paths starting with
+  // "ecas" or "ecas" followed by alphanumeric characters.
+  if (preg_match('/^ecas$|^ecas[a-zA-Z0-9]|^sitemap\./m', $path)) {
```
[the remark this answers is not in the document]
gillesdeudon: With this regex, I excluded any path statring with "ecas_" or a "ecas-" but I included patches that contains "ecas" followed by alphanumeric characters or "ecas".
I tested without `^ecas$` but the last case ("ecas") was not included.

profiles/common/modules/custom/ecas/includes/ecas.inc
gillesdeudon: I will find a better way.

I was not aware about the submit callback. I place the test there to let sites altering the process for there specific use cases. So, they can alter the platform behaviour without twisting the platform.

profiles/common/modules/custom/ecas/includes/ecas.inc
gillesdeudon: I will comment in the ticket for this part.

gillesdeudon:
After discussion with business and the concerned developer, I close this PR. The work will be done in NEPT-520.

PR #2142, opened 2018-08-09
NEPT-2021: Block EU login if the e-mail is not present in the profile

gillesdeudon:
### Description

Implement a control on the presence of the e-mail address in the EU login user profile.
If it is not there, the Drupal user creation is stopped and the user is redirected to an error page with information about how to unblock the situation.

### Change log

- Added:
  - profiles/common/modules/custom/ecas/includes/ecas.inc: Add control on the presence of the e-mail and the redirection process
  - profiles/common/modules/custom/ecas/ecas.variable.inc: Declaring the translatable variables
  - profiles/common/modules/custom/ecas/includes/ecas.admin.inc: Add 2 field in the admin form for the error messages to display; one for a EU Login issue and one when the issue is related to social network.
  - profiles/common/modules/custom/ecas/ecas.install and /ecas.install.inc: Set the default value for the 2 new variable at install time.
  - profiles/common/modules/custom/ecas/ecas.tokens.inc: Declare new ecas specific tokens, 5 allowing  to introduce default illustration images in the error page and 2 to include URLs linked to the EU Login interface.
- Changed:
  - /profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.callbacks.inc: Modify the language suffix rewrite callback to include the ecas warning and error pages in the suffix treatment and keep the other ecas page oustide.

### Commands

drush cc all

At site level, they must be notified that the ecas variable are translatable now if these variables are flagged as translatable in i18n_variable module.


gillesdeudon:
After discussion with business and the concerned developer, I close this PR. The work will be done in NEPT-520.

PR #2146, opened 2018-08-13
NEPT-1999: Schedule publishing of new version of node

enriquelacoma:
### Description

Users need to be able to schedule the publishing of a new version of an already published node.
### Change log

- Added: add patches to the deploy 
    .- patches/scheduler_workbench-allowed_status.patch
    .- https://www.drupal.org/files/issues/scheduler_workbench-revision_publish-2048999-45.patch


resources/multisite_drupal_standard.make
```diff
@@ -656,6 +656,13 @@ projects[scheduler][version] = 1.5
 
 projects[scheduler_workbench][subdir] = "contrib"
 projects[scheduler_workbench][version] = 1.3
+; Allow to schedule the publish date of a revision
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1999
+; https://www.drupal.org/project/scheduler_workbench/issues/2048999
+projects[scheduler_workbench][patch][] = https://www.drupal.org/files/issues/scheduler_workbench-revision_publish-2048999-45.patch
+; Allow publication of revisions with selected status. This patch requires the from issue https://www.drupal.org/project/scheduler_workbench/issues/2048999
```
imanoleguskiza: There is a type here, it should say "This patch requires the patch from issue "

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++ * @see system_settings_form()
++ */
++function scheduler_workbench_allowed_status_form($form, &$form_state) {
++  $options = workbench_moderation_state_labels();
++  $form['scheduler_workbench_allowed_status'] = array(
++    '#type' => 'checkboxes',
++    '#title' => t('Allowed Workbench Moderation status for scheduling publication'),
++    '#options' => $options,
++    '#default_value' => variable_get('scheduler_workbench_allowed_status', array()),
++    '#description' => t('Choose status to allow revisions for scheduling publication. When Scheduler Workbench tries to publish a revision i will check if the current state is one of the selected from this list.'),
```
imanoleguskiza: There is a type here, it should say "When Scheduler Workbench tries to publish a revision **it** will check"

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
+ /**
++ * Implements hook_menu().
++ */
++function scheduler_workbench_menu() {
++  $items['admin/config/content/scheduler_workbench'] = array(
++    'title' => 'Workbench Scheduler publication states',
++    'description' => 'Select the states that can be scheduled for publication',
++    'page callback' => 'drupal_get_form',
++    'page arguments' => array('scheduler_workbench_allowed_status_form'),
++    'access arguments' => array('administer scheduler workbench'),
```
imanoleguskiza: The permission name here differs from the one defined below: here its "administer scheduler workbench" but then we defined it as "scheduler workbench"

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++function scheduler_workbench_update_7001() {
++  if (!variable_set('scheduler_workbench_allowed_status', array())) {
++    _scheduler_workbench_status();
++  }
++}
++
++/**
++ * Sets the allowed status that can be scheduled for publication.
++ */
++function _scheduler_workbench_status() {
```
imanoleguskiza: The functione name here is a bit weird, it doesn't tell anything about what it does. I think a better name would be _scheduler_workbench_define_allowed_status

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++  $transitions = workbench_moderation_transitions();
++  $status = array();
++  foreach ($transitions as $key => $value) {
++    if ($value->to_name != 'published') {
++      unset($transitions[$key]);
```
imanoleguskiza: This interaction seems unnecessary, can't we simply add $value->from_name to the $status array if the condition is met?
`if ($value->to_name == 'published') {
  $status[] = $value->from_name;
}
`

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++  return $permissions;
+ }
+
+ /**
+@@ -252,6 +271,22 @@ function _scheduler_workbench_scheduler_unpublish(&$node) {
+  *   Node being published
+  */
+ function _scheduler_workbench_scheduler_pre_publish(&$node) {
++  // Get status allowed to be published.
++  $status = variable_get('scheduler_workbench_allowed_status', array());
```
imanoleguskiza: The variable name is misleading here, I think $allowed_status is much better ( same for other appearances of the same variable)

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++  foreach ($status as $key => $value) {
++    if ($value === 0) {
```
imanoleguskiza: In which situation will value equal 0?

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
++  // Check if the revision current status allows it to be published.
++  if (!in_array($node->workbench_moderation['my_revision']->state, $status)) {
++    $options = array(
++      "%node" => $node->nid,
++      '%status' => $node->workbench_moderation['my_revision']->state,
++    );
++    watchdog("scheduler_workbench", t("Node %node can't be published, check status allowed to be published."), $options, WATCHDOG_INFO);
++    return;
++  }
+   if ($state = variable_get('scheduler_publish_moderation_state_' . $node->type, FALSE)) {
```
imanoleguskiza: The code added above should probable be inside this condition, that way we don't do unnecessary checks if the content type doesn't have scheduler enabled.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,147 @@
...
+ function _scheduler_workbench_scheduler_pre_publish(&$node) {
++  // Get status allowed to be published.
++  $status = variable_get('scheduler_workbench_allowed_status', array());
++  foreach ($status as $key => $value) {
++    if ($value === 0) {
++      unset($status[$key]);
++    }
++  }
++  // Check if the revision current status allows it to be published.
++  if (!in_array($node->workbench_moderation['my_revision']->state, $status)) {
```
imanoleguskiza: The check here is fine but I think we shouldn't allow the users to define a scheduling date if the revision state is not allowed to begin with.
I would do this on scheduler_workbench_form_alter, just before calling scheduler_workbench_add_revision_publish_on_field and I would add a message to inform the user that publishing the revision in that state is not allowed.

imanoleguskiza (changes requested):

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: This allows, for example, to schedule the publication of a draft and publish it only if it is in Validation, that way you can schedule a publication and publish it only if it is in a correct status.

If a revision is not in a valid status, i will not be published and register a watchdog message with the error.

I think that we could keep the current funtionality and, as you suggest, add a message informing the user that it will no be published with the current status.

tests/src/Context/SchedulerContext.php
```diff
@@ -39,4 +58,33 @@ public function iChangeUnpublishingDateOfTypeNodeWithTitleToDate($type, $title,
     }
   }
 
+  /**
+   * Adds the current day with format Y-m-d.
+   *
+   * @Then I fill in :arg1 with current day
+   */
+  public function iFillInWithCurrentDay($arg1) {
```
imanoleguskiza, line 66: This two steps seem to be unused so I guess we should delete them.

tests/features/scheduler.feature
```diff
@@ -10,28 +10,32 @@ Feature: Scheduler features
   Scenario: User can schedule a date to publish a content
     When I go to "node/add/page"
```
imanoleguskiza, line 41: The tests are fine but we are missing the tests where we publish a new revision of an existing node. We also need a tests that shows how a node with an incorrect state is not published when the date comes.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,173 @@
...
++      }
++    }
++    // Check if the revision current status allows it to be published.
++    if (!in_array($node->workbench_moderation['my_revision']->state, $allowed_status)) {
++      $options = array(
++        "%node" => $node->nid,
++        '%status' => $node->workbench_moderation['my_revision']->state,
++      );
++      watchdog("scheduler_workbench", t("Node %node can't be published, check status allowed to be published."), $options, WATCHDOG_INFO);
++      return;
```
imanoleguskiza, line 163: In my tests the watchdog was triggered but the scheduling proceeded as usual so I think the check needs to be triggered somewhere else.

imanoleguskiza (changes requested):

tests/src/Context/SchedulerContext.php
enriquelacoma: We use them now in the scenario to check the publication of revision

tests/features/scheduler.feature
enriquelacoma: Added two new scenarios

AlbaParrilla (dismissed):

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,206 @@
...
++ * Implements hook_uninstall().
++ */
++function scheduler_workbench_uninstall() {
++  variable_del('scheduler_workbench_allowed_status');
++}
++
++/**
++ * Implements hook_update_N().
++ */
++function scheduler_workbench_update_7001() {
```
gillesdeudon: We should not have this new hook_update in 2.5. I suggest to remove it from the "local" patch.

resources/multisite_drupal_standard.make
```diff
@@ -653,6 +653,15 @@ projects[scheduler][version] = 1.5
...
 projects[scheduler_workbench][subdir] = "contrib"
 projects[scheduler_workbench][version] = 1.3
+; Allow to schedule the publish date of a revision
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1999
+; https://www.drupal.org/project/scheduler_workbench/issues/2048999
+projects[scheduler_workbench][patch][] = https://www.drupal.org/files/issues/scheduler_workbench-revision_publish-2048999-45.patch
+; Allow publication of revisions with selected status. This patch requires the patch from issue https://www.drupal.org/project/scheduler_workbench/issues/2048999
+; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1999
+; This patch requieres the patch https://www.drupal.org/files/issues/scheduler_workbench-revision_publish-2048999-45.patch from
+; this issue https://www.drupal.org/project/scheduler_workbench/issues/2048999.
```
gillesdeudon, line 677: We need a reference to the D.O issue from where this "local" patch id derived like that we can ensure the traceability of the patch and ease our job during future module upgrades.

gillesdeudon (changes requested):
@enriquelacoma could you tackle the comments I made?

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
+--- a/scheduler_workbench.module
++++ b/scheduler_workbench.module
+@@ -7,6 +7,22 @@
+  */
+
+ /**
++ * Implements hook_menu().
++ */
++function scheduler_workbench_menu() {
++  $items['admin/config/content/scheduler_workbench'] = array(
```
gillesdeudon: IMO, this menu time should be either a child of "Scheduler" item (admin/config/scheduler) or of "Workbench moderation" item (admin/config/workbench/moderation) in order to be consistent with the other contrib modules that enrich Scheduler or Workbench moderation like "Workbench moderation" does with the "Workbench" module. 

I am more in favour of "Scheduler" item with in the `scheduler_workbench_allowed_status_form()`, a mention of the Workbench moderation transitions and the admin form managing them.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
+
+ /**
+@@ -141,7 +160,14 @@ function scheduler_workbench_scheduler_node_form_submit($form, $form_state) {
+  */
+ function scheduler_workbench_node_validate($node, $form) {
+   $date_format = variable_get('scheduler_date_format', SCHEDULER_DATE_FORMAT);
+-
++  if (!empty($node->revision_publish_on) || !empty($node->publish_on) ) {
++    // Check if the current status will be published.
++    $allowed_status = variable_get('scheduler_workbench_allowed_status', array());
```
gillesdeudon: IMO, if `scheduler_workbench_allowed_status`is not set, we should come back to the transition settings and only give the list of states that can be lead to "published"; instead of an empty array.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
++    drupal_set_message($message, 'status', FALSE);
+   }
+ }
+
+@@ -253,6 +280,23 @@ function _scheduler_workbench_scheduler_unpublish(&$node) {
+  */
+ function _scheduler_workbench_scheduler_pre_publish(&$node) {
+   if ($state = variable_get('scheduler_publish_moderation_state_' . $node->type, FALSE)) {
++    // Get status allowed to be published.
++    $allowed_status = variable_get('scheduler_workbench_allowed_status', array());
```
gillesdeudon: Same comment as I did for `scheduler_workbench_node_validate()`.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
++    }
+     if (!empty($node->workbench_moderation['my_revision']) &&
+       $node->workbench_moderation['my_revision']->state == $state
+     ) {
+@@ -319,6 +363,15 @@ function scheduler_workbench_form_workbench_moderation_moderate_form_alter(&$for
+ function scheduler_workbench_revision_publish($current, $state) {
+   workbench_moderation_moderate($current, $state);
+   $create_publishing_revision = variable_get('scheduler_publish_revision_' . $current->type, 0) == 1;
++  $allowed_status = variable_get('scheduler_workbench_allowed_status', array());
++  if (!in_array($current->workbench_moderation['my_revision']->state, $allowed_status)) {
```
gillesdeudon, line 181: Same remark as I made for `scheduler_workbench_node_validate()`. Maybe encapsulating this variable_get call in its own function could be interesting.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
+@@ -0,0 +1,18 @@
++<?php
++
++/**
++ * Form callback for the settings form.
++ *
++ * @see system_settings_form()
++ */
++function scheduler_workbench_allowed_status_form($form, &$form_state) {
++  $options = workbench_moderation_state_labels();
```
gillesdeudon, line 38: We should tie the options on the Workbench moderation transitions possibilities like done in `_scheduler_workbench_define_allowed_status()`. 
That means this function should move to the module file. I think also it could be integrated in the module API and become public.

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
++  $form['scheduler_workbench_allowed_status'] = array(
++    '#type' => 'checkboxes',
++    '#title' => t('Allowed Workbench Moderation status for scheduling publication'),
++    '#options' => $options,
++    '#default_value' => variable_get('scheduler_workbench_allowed_status', array()),
++    '#description' => t('Choose status to allow revisions for scheduling publication. When Scheduler Workbench tries to publish a revision it will check if the current state is one of the selected from this list.'),
```
gillesdeudon: In link with my comment above, we could add in the description a text referring to the "Transition" interface of Workbench moderation to add more Moderate state.

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Function removed

resources/multisite_drupal_standard.make
enriquelacoma: Comment updated

resources/patches/scheduler_workbench-allowed_status.patch
```diff
@@ -0,0 +1,197 @@
...
++  }
+   if ($create_publishing_revision) {
+     $current->revision = TRUE;
+     $current->log = t('Saved by Scheduler workbench on @now.', array(
+@@ -348,6 +401,16 @@ function scheduler_workbench_scheduler_allow_publishing($node) {
+     // If this state is one that the current user can access, return TRUE.
+     if ($permitted_states = workbench_moderation_states_next($current_state, $draft_owner, $node)) {
+       $allowed = isset($permitted_states[$published_state]);
++    } // Get status allowed to be published.
++    $allowed_status = variable_get('scheduler_workbench_allowed_status', array());
```
gillesdeudon: Same remark as I made for scheduler_workbench_node_validate(). Maybe encapsulating this variable_get call in its own function could be interesting.

gillesdeudon (changes requested):
I leave some comments to address before being able to merge.

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Created wrapper _scheduler_workbench_allowed_status()

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Updated description to include default values

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Included as a scheduler tab

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Created wrapper _scheduler_workbench_allowed_status()

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Created wrapper _scheduler_workbench_allowed_status()

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Created wrapper _scheduler_workbench_allowed_status()

resources/patches/scheduler_workbench-allowed_status.patch
enriquelacoma: Created wrapper _scheduler_workbench_allowed_status()

ventzie555 (dismissed):
I tested it and everything appears to work fine. Code also look good to me.

imanoleguskiza (dismissed):

gillesdeudon (dismissed):

gillesdeudon:
I approve in order to unblock the PR as it is in a new code review process

AlbaParrilla (approved):
I just changed a message a little bit, the rest is working properly. 
Ok for me

Fefaine (approved):
OK

Fefaine:
🎆

PR #2169, opened 2018-08-30
NEPT-2047: Fasttrack, use the product identifier 'EDT' rather than 'REV'

aleach:
### Description

Update fast track translation to use the product identifier 'EDT' rather than 'REV'.

### Change log

- Added: Custom product identifier within the review request method to pass 'EDT'.
- Changed: oe-poetry-client to version 0.3.6 which supports custom product identifier's. Additional test coverage of the product identifier within the fast track tests.

### Commands

Not applicable.

### DIFF
--- a/tests/features/tmgmt/fast-track.feature
+++ b/tests/features/tmgmt/fast-track.feature
@@ -118,7 +118,8 @@ Feature: Fast track
         "DO" : [
           { "ne_dgt_rules_ftt_node_send_translation_request" : {
               "USING" : { "node" : [ "node" ],
-               "delay" : "2017-12-01 15:00:00"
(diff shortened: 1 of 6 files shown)


tests/features/tmgmt/fast-track.feature
```diff
@@ -148,6 +148,7 @@ Feature: Fast track
     And I press "Apply"
     Then I should see "Revision state: Validated"
     And Poetry service received request should contain the following text:
+      | <produit>EDT</produit>                                        |
```
gillesdeudon: The test of the "produit" value should be done after transition from "draft to "Needs review" above after the line 146. Here I propose to test on `<produit>TRA</produit>`

PR #2186, opened 2018-09-17
NEPT-2029: Update TMGMT patch for better control of translated data length

dxvargas:
### Description

Update TMGMT patch for better control of translated data length when field has columns (multi values).

### Change log

- Changed: TMGMT patch updated.

### Commands

No commands needed.


ventzie555 (approved):

enriquelacoma (approved):

resources/multisite_drupal_standard.make
```diff
@@ -739,7 +739,8 @@ projects[tmgmt][patch][] = https://www.drupal.org/files/issues/2812863.patch
 ; #2362321 : Check source length limits
 ; https://www.drupal.org/node/2362321
 ; https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-1802
```
ku-enza, line 741: Hi,
Maybe you can also remove this line except if the new patch covers both issues.

PR #2209, opened 2018-10-08
NEPT-794: Customizable message in workbench moderation is broken

AlbaParrilla:
### Description

When I create a new multilingual content and I click save I should see a warning message in just one line, but the token of the title of the content adds a "p" html tag that breaks the message into 3 lines. 
The warning message must be something like this: "The state of the content Title in English and all its validated translations English French will be updated!"

### Change log

- Changed: nexteuropa_multilingual.module (filtered html tags in the token of the title to let only bold and italics tags)

### Commands

No commands needed


profiles/common/modules/features/nexteuropa_multilingual/nexteuropa_multilingual.module
```diff
@@ -370,7 +370,7 @@ function _nexteuropa_multilingual_warning_message(&$form, $node) {
   }
   $warning_message_languages = variable_get('nexteuropa_multilingual_warning_message_languages');
   if (!empty($warning_message_languages)) {
-    $message = token_replace($warning_message_languages, array('node' => $node));
+    $message = strip_tags(token_replace($warning_message_languages, array('node' => $node)), '<b><i><strong><em>');
```
aritomelo: Is this list of allowed tags come from somewhere and it is self-decision?

It is just to know and I can approve the PR as it is!

AlbaParrilla: I don't know if there is a rule for the platform that tells which tags can be used in this kind messages. I looked for different messages in the platform and I realised that only italics and bold tags were used in the messages that I found, so those are the ones I added here.

aritomelo: Thx @AlbaParrilla  for your explanation.

Ticket ready for next step.

PR #2214, opened 2018-10-09
NEPT-336: Add composer vulnerability check

jfhovinne:
### Description

Ensure we do not use Composer packages marked as vulnerable.

### Change log

- Added: Composer packages check, using Roave security advisories.

### Commands

No command.


ventzie555 (approved):
I tested it and it stops the composer install execution if a package version is in conflict with roave/security-advisories package. It woks as expected.

ku-enza (approved):
Tested with a vulnerable package version.
It's OK, roave dependency is fired properly

jfhovinne:
Replaced by PR 2257.

PR #2225, opened 2018-10-15
NEPT-572: Move dependencies from profile to modules.

Fefaine:
### Description

Move dependencies from profile to modules.

### Change log

- Changed: Dependencies to diff added to modules are removed from profiles


ventzie555 (approved):
I tested it and the code looks good.

aritomelo (approved):
Ready to go.

PR #2242, opened 2018-10-23
Nept 1606 jenkins drone comparrison

aleach:
DO NOT MERGE.


aleach:
Verified Jenkins is also passing the test's against the code.

PR #2244, opened 2018-10-24
NEPT-2111: Remove timezone override on mpdf library

enriquelacoma:
### Description
 The library mpdf has a bug that overrides the platform timezone, as the version 6 of mpdf is fronzen, we created a local patch to remove the override.

### Change log

- Added: patch resources/patches/mpdf_timezone.patch


gillesdeudon (approved):
Ok for me

Fefaine (approved):
OK

PR #2247, opened 2018-10-26
Feature: Update existing users with the same email

XenoVkl:
### Description

Consider the following use case scenario:
A user has registered on the drupal website (not via ecas)
At some point, the user attempts to use the the EU login and enters the same email with the one that is registered on drupal.
Once authenticated, a new account will be created for the user instead of syncing the info with the existing one (assuming that drupal and ecas usernames are different).
That is happening because in the _ecas_login_check()_ function, users are loaded by their username. 
I suggest that in case _user_load_by_name()_ fails, we try loading the user by mail and if that succeeds, register the user to the authmap table and update the username.

### Change log
Changed _profiles/common/modules/custom/ecas/includes/ecas.inc_ in order to provide a solution for the above use case scenario

### DIFF
--- a/profiles/common/modules/custom/ecas/includes/ecas.inc
+++ b/profiles/common/modules/custom/ecas/includes/ecas.inc
@@ -397,6 +397,13 @@ function ecas_login_check() {
       // @codingStandardsIgnoreEnd
       // Try to log into Drupal.
       $account = user_load_by_name($ecas_name);
+      
+      if (!$account) {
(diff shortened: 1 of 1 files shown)


cherouvim:
This is useful. Note that until this is merged, it's possible that duplicate accounts with the same email are stored within Drupal, which is something that would break things.

cherouvim:
No merge?

PR #2248, opened 2018-10-29
NEPT-1736: Update mpdf library to version 7

enriquelacoma:
### Description

Update the version of the pdf library to 7

### Change log

- Added: Composer install command to phing
- Removed: patch resources/patches/mpdf_timezone.patch


build.package.xml
```diff
@@ -67,6 +67,10 @@
             <include name="*.*" />
           </fileset>
         </delete>
+        <composer command="install" composer="${composer.bin}">
```
ku-enza, line 70: I am not confident with these lines, it's kind of double work as the make file manage to do it.

ventzie555 (approved):
I did the QA, checked the code and everything looks good to me.

PR #2249, opened 2018-10-29
NEPT-2092: Target build-platform-dev seems to not clean properly the contrib folder inside the profile

enriquelacoma:
### Description

Delete contrib module folder from profiles when building the platform

### Change log

- Added: phing command to delete modules/contrib module

### Commands

Build platform ./bin/phing build-platform-dev

### DIFF
--- a/build.package.xml
+++ b/build.package.xml
@@ -19,6 +19,8 @@
 
         <echo msg="Delete previous build." />
         <delete dir="${phing.project.build.dir}" includeemptydirs="true" failonerror="false" />
+        <echo msg="Delete previous contrib folder. profiles/${platform.profile.name}/modules/contrib" />
+        <delete dir="profiles/${platform.profile.name}/modules/contrib" includeemptydirs="true" failonerror="false" />
     </target>
 
     <!-- Make Drupal core. -->
--- a/build.package.xml
+++ b/build.package.xml
@@ -19,8 +19,18 @@
 
         <echo msg="Delete previous build." />
         <delete dir="${phing.project.build.dir}" includeemptydirs="true" failonerror="false" />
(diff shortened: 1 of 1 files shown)


ventzie555 (approved):
Everything looks good.

ventzie555 (changes requested):
For some reason I am getting an error when I try to do the spinup:

Unable to copy /tmp/make_tmp_1541499582_5be16abe9e5b3/__build__/profiles/multisite_drupal_standard to /home/ec2-user/environment/sites/2092/build/profiles/multisite_drupal_standard.                                      [error]
Cannot move build into place.                                                                                                                                                                                              [error]

BUILD FAILED
/home/ec2-user/environment/sites/2092/build.package.xml:56:41: Task exited with code 1

It happens on make-platform

ventzie555 (dismissed):
It looks like the issue I had was due to running out of space on my disc

Fefaine (approved):
Ok for me !

Fefaine (approved):
OK

ventzie555 (changes requested):
I'm not sure about this but should we also delete the libraries folder?
Here is a paragraph from .gitignore:

profiles/multisite_drupal_*/libraries
profiles/multisite_drupal_*/modules/contrib
profiles/multisite_drupal_*/themes

Not sure about this also but I guess option is to use overwirte:

"Allows the project to be installed in a directory that is not empty. If not specified this is treated as FALSE, drush_make sets an error when the directory is not empty. If specified TRUE, drush_make will continue and use the existing directory. Useful when adding extra files and folders to existing folders in libraries or module extensions.

    projects:
      myproject:
        overwrite: TRUE
"

ventzie555 (approved):
Since it is not possible to fix the issue with libraries folder, I am approving it.

PR #2256, opened 2018-11-05
Nept 2125: events_og_update_7101 generates an error when context does not exist

Fefaine:
### Description

events_og_update_7101 generates an error when context does not exist

### Change log

- Added: additional check on toolbox functions


aleach (approved):
Looks good. I do prefer exiting functions quickly with the reverse check though
`
if (!$context) {
  return;
}
// Code
`
Rather than
`
if ($context) {
  // Code
}
`
But in this case, the code isn't that complicated to read.

ku-enza (approved):
It's OK for me.

PR #2257, opened 2018-11-05
NEPT-336: Use Roave advisories

jfhovinne:
### Description

Ensure we do not use Composer packages marked as vulnerable.

### Change log

- Added: Composer packages check, using Roave security advisories.

### Commands

No command.


aleach (approved):
+1 Checked out, ran composer install, then tried to require packages outlined in the conflicts region of the composer json.

Fefaine (approved):
OK

PR #2258, opened 2018-11-05
NEPT-1730: PHP version dependency

aleach:
### Description

Ensure the specification of a requirement of PHP 5.6 or above.

### Change log

- Changed: Modules that contained the 'use function' function importing technique to directly require a minimum of php 5.6. Update the Readme and composer requirements.

### Commands

Not applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 * Composer
 * PHP Phar extension
 * PhantomJS (in order to run JavaScript during Behat tests)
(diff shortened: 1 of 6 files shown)


ventzie555 (approved):
Code looks good and when tested with php <= 5.6 error is raised.

aritomelo (approved):
The solution is OK, tested ok.

PR #2262, opened 2018-11-07
NEPT-570: Remove facetapi from profile dependencies

ventzie555:
### Description

Remove facetapi from profile dependencies

### Change log
- Changed:   profiles/multisite_drupal_communities/multisite_drupal_communities.info
- Changed:   profiles/multisite_drupal_standard/multisite_drupal_standard.info


Fefaine (changes requested):
The feature multisite settings core adds the current_search block in the sidebar left of the search context.
I can't understand why. I suggest we move this block insertion inside the solr_config feature
THat's the only feature that requires facetapi, and current_search is a submodule of it.

Fefaine (approved):

aritomelo (approved):
After the bellows steps I give +1 to the PR.

- Confirmed that faceapi was installed/enabled by default
- Check if there is any remaning facetapi reference outside solr_config
- create a new instance based on branch nept-570-release-2.6
- Check if the module has been enabled by default, as expected NOT enabled.
- enabled the solr_config
- Recheck if the module has been enabled by enabling Solr config, and as expected It enables

AlbaParrilla (approved):
Ok for me 👍

PR #2264, opened 2018-11-07
NEPT-1542: Add the possibility to remove only Varnish cache

Fefaine:
### Description

Add the possibility to remove only Varnish cache

### Change log

- Changed: Varnish menu items, to launch only drupal cache clear when checkbox is ticked

### Commands

`drush cc all`


AlbaParrilla (approved):
👍

aritomelo (approved):
The solution looks great and user test shows that there is 2way to clean the cache.

PR #2274, opened 2018-11-15
NEPT-127: Remove piwik contrib module, we do not use it.

Fefaine:
### Description

Remove piwik contrib module

### Change log

- Removed: Remove piwik contrib module


AlbaParrilla (approved):
Ok for me

PR #2277, opened 2018-11-16
NEPT-2137: Restore dependency declaration to ex-tidy

AlbaParrilla:
### Description

ext-tidy dependency declaration was removed from the composer.json file by mistake and it has to be restored

### Change log

- Changed: composer.json and composer.lock

### Commands

No commands


aritomelo (approved):
Ready to go.

AlbaParrilla (commented):
The comment about Poetry that should be added, as it is in acceptance criteria, could not be added as json files cannot have comments.

ku-enza (approved):
Ok with this

Fefaine (approved):
ok for me

PR #2278, opened 2018-11-17
NEPT-1259: Apply the fix to both profiles.

Fefaine:
### Description

Apply the mpdf fix to both profiles.

### Change log

- Changed: hard coded profile in build.package.xml


ku-enza (approved):
It's better.

PR #2279, opened 2018-11-19
NEPT-573: remove rate from the profile dependencies.

Fefaine:
## NEPT-573

Remove rate from the profile dependencies

### Change log

- Removed: code that forces rate throug the profile and cce_basic_config


aritomelo (approved):
Looks ok.

PR #2284, opened 2018-11-19
NEPT-2148: Check if the destination url is not set otherwise redirect

Fefaine:
### Description

Check if the destination url is not set otherwise redirect to user profile

### Change log

- Changed:

Function _ecas_logged_user_profile_redirect


AlbaParrilla (approved):
Ok for me 👍

aritomelo (approved):
After testing the Solution on c9, I was able to use EU-login and been redirect to the right page, which it wasn't possible previous.

So the PR can progress.

PR #2285, opened 2018-11-20
NEPT-664: Remove fpa from platform

AlbaParrilla:
### Description

Remove fpa from platform. The module should not be used on production since all permissions should be exported.

### Change log

- Changed: multisite_drupal_communities.info, multisite_drupal_standard.info, multisite_drupal_standard.make

### Commands

No commands


aritomelo (approved):
Ready to go!

ku-enza (approved):
Done

PR #2287, opened 2018-11-21
NEPT-2152: Issue with scheduler, check variable exists

enriquelacoma:
### Description

Add check to ensure variable exist

### Change log

- Changed: Update patch scheduler_workbench-allowed_status.patch


Fefaine (approved):
OK

aritomelo (approved):
From new instance it seems no regression, so I give my +1

PR #2289, opened 2018-11-22
NEPT-2155: Remove useless module, already called in profile standard

Fefaine:
### Description

Remove second download of the nexteuropa_poetry module.

### Change log

- Changed:
resources/multisite_drupal_communities.make file


AlbaParrilla (approved):
Ok for me

AlbaParrilla:
Sorry, I didn't see that Jenkins had failed. I will give my approval once Jenkins is green

aritomelo (approved):
After performance the follow steps, I give my +1 and the ticket can progress.

Check that the next_europa_poetry is reference from multisite_drupal_standard and removed from communities
Install the new instance from the PR branch
Check if the module has been downloaded
also check if it is on the right folder, which means to be in contrib as the rule followed by platform projects.

PR #2297, opened 2018-11-26
  NEPT-2029: Unable to request translation "Data too long" error

aritomelo:
### Description
WIP: TEMPORARLLY PATCH applied, only for CODE REVIEW before it goes to dOrg.
Unable to request translation "Data too long" error

### Change log
- Fixed:
    - Fixed the bug for body field

    - Implemented the logic to treat text_with_sumary and long_text field type, which was the the bug cause.

    - Implemented a logic for link test.

    - Break the whole logic into two function.

    - Small improvements and typo.

### DIFF
--- a/tests/features/menus/feature_set.feature
+++ b/tests/features/menus/feature_set.feature
@@ -75,7 +75,7 @@ Feature: Feature set menu
     Then I should not see the text "F.A.Q"
     Then I should not see the text "Press Release"
     Then I should not see the text "Site activity"
-    Then I should not see the text "Maxlength"
+#    Then I should not see the text "Maxlength"
     Then I should not see the text "News"
     Then I should not see the text "Newsletters"
     Then I should not see the text "GIS field"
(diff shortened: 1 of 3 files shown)


tests/features/menus/feature_set.feature
```diff
@@ -75,7 +75,7 @@ Feature: Feature set menu
     Then I should not see the text "F.A.Q"
     Then I should not see the text "Press Release"
     Then I should not see the text "Site activity"
-    Then I should not see the text "Maxlength"
+#    Then I should not see the text "Maxlength"
```
ku-enza: Hi,
This seems unnecessary for me as this test is not about translations.
The PR should contains  changes for the issue only.
There is no reason that the text 'Maxlength' appears for this user because of your patch.
Seems like a CI bug.

ku-enza (changes requested):
Chan you un-comment the line in the behat test and see if Jenkins is OK with that.

ku-enza (approved):

Fefaine (approved):
OK for me

PR #2304, opened 2018-12-05
NEPT-762: Remove date_api_version variable from cce_basic_config

AlbaParrilla:
### Description

This parameter must be removed from cce_basic_config because it is set automotically by the "date_api" module and that must not set like that.

### Change log

- Removed: date_api_version variable from the following files:
  - profiles/common/modules/features/cce_basic_config/cce_basic_config.info
  - profiles/common/modules/features/cce_basic_config/cce_basic_config.strongarm.inc

### Commands

drush fr cce_basic_config


Fefaine (approved):
OK

PR #2306, opened 2018-12-07
Update Release 2.6 with Release 2.5

Fefaine:
### Description

Check if destination URL is not set, or redirest


aritomelo (approved):
I have not test the platform, however, for the changes and the green from CI I give my +1.

PR #2312, opened 2018-12-12
Nept 2185: CKEditor height does not reflect the rows attribute

aritomelo:
### Description

The issue: Textarea with different height (Default height of free text fields)
The patch is already provided by drupal community.

### Change log

- Added: patch on wysiwy js editor files to force an approximate size.


AlbaParrilla (approved):

ku-enza (approved):
Patch fix the issue..
Ok

PR #2317, opened 2018-12-14
NEPT-2149: CLONE - NextEuropa_token_CKEditor - Ajax Exposed filters not working with multiple same views.

AlbaParrilla:
### Description

The problem comes from the fact that we have several CKEditor on the page.
Clicking the icon to add a link, opens the popup each time with the same view (duplicated).
The view of the nexteuropa_token_ckeditor module uses AJAX with better exposed filter. 
The JS behaviors of the view module are based on the name of the view to initialize the ajax actions. 
Since there are several times this same view on the page, the javascript selector is always the same and this creates a bug with the AJAX exposed filter between all the views.

### Change log

- Added: patch for views added in resources/multisite_drupal_standard.make


ku-enza (approved):
The patch fixed the issue.

Fefaine (approved):
OK!

PR #2318, opened 2018-12-17
NEPT-2183: Fix versions and remove warnings on overlap

Fefaine:
### Description

Fix versions and remove warnings on overlap

### Change log

- Added: regex tests
- Fixed: Test on rules using regex


aritomelo (approved):
The fix works , so ready to be merged.

ku-enza (approved):
Approved.

PR #2321, opened 2018-12-20
NEPT-2178: Translation workflow may not ensure that text removed from original is also removed from translations.

aritomelo:
### Description

Translation not taking in account the update on fields when the field updated has been empty after been populated.

### Change log

- Added:
   - A patch has been created.
   - An issue open on dorg will follow the patch. (tbd)
   - The solution target the translation using File and auto_accept and also the small jobs.

### Commands

n/a

### DIFF
--- a/profiles/common/modules/custom/tmgmt_workbench/includes/tmgmt_workbench.plugin.inc
+++ b/profiles/common/modules/custom/tmgmt_workbench/includes/tmgmt_workbench.plugin.inc
@@ -52,9 +52,9 @@ class TMGMTWorkbenchSourcePluginController extends TMGMTEntitySourcePluginContro
     $handler = entity_translation_get_handler($job_item->item_type, $current_revision);
     $handler->setActiveLanguage($job_item->getJob()->target_language);
 
-    $current_revision_metadata_value = $this->manageEmptyFieldUpdate($current_revision, $job_item->getData(), $job->source_language, $job->target_language);
-    if ($current_revision_metadata_value) {
-      $current_revision = $current_revision_metadata_value;
+    $current_revision_update = $this->manageEmptyFieldUpdate($current_revision, $job_item->getData(), $job->source_language, $job->target_language);
+    if ($current_revision_update) {
+      $current_revision = $current_revision_update;
     }
     // Save current node revision.
     node_save($current_revision);
--- a/profiles/common/modules/custom/tmgmt_workbench/includes/tmgmt_workbench.plugin.inc
+++ b/profiles/common/modules/custom/tmgmt_workbench/includes/tmgmt_workbench.plugin.inc
@@ -52,6 +52,10 @@ class TMGMTWorkbenchSourcePluginController extends TMGMTEntitySourcePluginContro
     $handler = entity_translation_get_handler($job_item->item_type, $current_revision);
     $handler->setActiveLanguage($job_item->getJob()->target_language);
 
+    $current_revision_metadata_value = $this->manageEmptyFieldUpdate($current_revision, $job_item->getData(), $job->source_language, $job->target_language);
+    if ($current_revision_metadata_value) {
+      $current_revision = $current_revision_metadata_value;
+    }
     // Save current node revision.
     node_save($current_revision);
 
--- a/resources/patches/translation_not_taking_in_account_the_source_data_update.patch
+++ b/resources/patches/translation_not_taking_in_account_the_source_data_update.patch
@@ -1,0 +1,62 @@
+diff --git a/sources/entity/tmgmt_entity.plugin.inc b/sources/entity/tmgmt_entity.plugin.inc
+index 90cf76e..66a0b89 100644
+--- a/sources/entity/tmgmt_entity.plugin.inc
++++ b/sources/entity/tmgmt_entity.plugin.inc
+@@ -49,6 +49,12 @@ class TMGMTEntitySourcePluginController extends TMGMTDefaultSourcePluginControll
+     $handler = entity_translation_get_handler($job_item->item_type, $entity);
+     $handler->setFormLanguage($job_item->getJob()->target_language);
+ 
++    // Copy over because if entity is not update, $entity_wrapper->save() won't work.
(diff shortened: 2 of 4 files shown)


profiles/common/modules/custom/tmgmt_workbench/includes/tmgmt_workbench.plugin.inc
```diff
@@ -52,6 +52,10 @@ class TMGMTWorkbenchSourcePluginController extends TMGMTEntitySourcePluginContro
     $handler = entity_translation_get_handler($job_item->item_type, $current_revision);
     $handler->setActiveLanguage($job_item->getJob()->target_language);
 
+    $current_revision_metadata_value = $this->manageEmptyFieldUpdate($current_revision, $job_item->getData(), $job->source_language, $job->target_language);
+    if ($current_revision_metadata_value) {
+      $current_revision = $current_revision_metadata_value;
+    }
```
dxvargas, line 58: This is our custom code so changes here should be covered with a behat test.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
+diff --git a/sources/entity/tmgmt_entity.plugin.inc b/sources/entity/tmgmt_entity.plugin.inc
+index 90cf76e..66a0b89 100644
+--- a/sources/entity/tmgmt_entity.plugin.inc
++++ b/sources/entity/tmgmt_entity.plugin.inc
+@@ -49,6 +49,12 @@ class TMGMTEntitySourcePluginController extends TMGMTDefaultSourcePluginControll
+     $handler = entity_translation_get_handler($job_item->item_type, $entity);
+     $handler->setFormLanguage($job_item->getJob()->target_language);
+ 
++    // Copy over because if entity is not update, $entity_wrapper->save() won't work.
```
dxvargas: This comment adds more confusion that understanding. entity_save is the standard way to update entities so there is no need to justify it's use.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
++    $entity_metadata_value = $this->manageEmptyFieldUpdate($entity, $job_item->getData(), $job->source_language, $job->target_language);
```
dxvargas: I would name the variable something like $entity_updated, to be more clear that it's the return of the manageEmptyFieldUpdate method.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
+     entity_save($job_item->item_type, $entity);
+ 
+     $translation = array(
+@@ -107,4 +113,44 @@ class TMGMTEntitySourcePluginController extends TMGMTDefaultSourcePluginControll
+ 
+     return array();
+   }
++
++  /**
++   * A controller method to check when the original entity has been updated.
```
dxvargas: My proposal:
Empty entity fields that come empty from translation.

At least, please avoid "controller" word, since it is usually used for PHP patterns like MVC etc.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
++   *
++   * In this case, if the field has been empty, the translation process does not take the update field,
++   * however, while saving the translated entity the updated field should be taken into account.
```
dxvargas: My proposal:

If a field comes empty from translation, it must be also processed, to be updated to empty.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
++   * @param $entity
++   *   The entity.
++   * @param $data_to_translate
++   *   A structured data array.
++   * @param string $source_language
++   *   The original language code.
++   * @param string $target_language
++   *   The language code to be translated to.
++   *
++   * @return The value of the wrapped data. If the data property is not altered, NULL
```
dxvargas: Please add the parameters types.

  * @param object $entity
  *   The entity.
  * @param array $data_to_translate
  *   A structured data array.
  * @param string $source_language
  *   The original language code.
  * @param string $target_language
  *   The language code to be translated to.
  *
  * @return object|null The value of the wrapped data. If the data property is not altered, NULL

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
++   *    is returned.
++   */
++  public function manageEmptyFieldUpdate($entity, array $data_to_translate, $source_language, $target_language) {
```
dxvargas: Please add the parameters types.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
```diff
@@ -0,0 +1,62 @@
...
++    // Filter the fields that are translated, and loop to the rest.
++    $cache = _field_info_field_cache();
++    $info = $cache->getBundleInstances($entity->entity_type, $entity->type);
```
dxvargas: Why are you using cache data?
There should be a better way. Method getData may be usefull, not sure.

dxvargas (dismissed):

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
aritomelo: thx mate, I will address your comments.

resources/patches/translation_not_taking_in_account_the_source_data_update.patch
aritomelo: No, getData only give you translated data, in this case it is all entity fields, but the purpose it is to get rid off fieds such as: vid, item_id, etc.

This part has been inspired from field project, on field_info_instance function, which you can find at field.info.inc line 627. The field module has many reference to it to get fields info.

enriquelacoma (approved):

ku-enza (approved):
Ok for me

PR #2328, opened 2019-01-09
NEPT-2204: For 2.6 release of platform update date module version and remove expired patches.

sergepavle:
### Description
This version implement PHP7 compatibility
However we noticed some hook update, so for 2.5 we need to add only the php 7 patches.
The real module upgrade will take place in 2.6

http://cgit.drupalcode.org/date/diff/?id2=7.x-2.10&id=7.x-2.11-beta1
### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:


Fefaine (approved):
OK

PR #2330, opened 2019-01-11
NEPT-2162: Filter the views in which title trimm is applied.

AlbaParrilla:
### Description

The title trim for long titles has to be done only in the views for the ckeditor filter views in order to avoid it to trim titles in other views in front-end. 

### Change log

- Changed: profiles/common/modules/custom/nexteuropa_token/modules/nexteuropa_token_ckeditor/nexteuropa_token_ckeditor.module

### Commands

No commands needed


ku-enza (approved):
Code is better it is sibling only the ckeditor view token content

Fefaine (approved):
OK for me

PR #2334, opened 2019-01-17
Revert "NEPT-2178: Translation workflow may not ensure that text removed from original is also removed from translations."

Fefaine:
Reverts ec-europa/platform-dev#2321


AlbaParrilla (approved):
Ok for me

enriquelacoma (approved):

PR #2337, opened 2019-01-17
NEPT-2214: Upgrade core to 7.63

Fefaine:
### Description

Upgrade core to 7.63

### Change log

- Changed: Drupal core version


Fefaine:
We have one branch "release-2.3" and one branch "support-2.3"
Find out what is the difference and which is deployed. 
Delete the other one or document better

Fefaine:
Gilles confirmed release-2.3 is the right branch to use

PR #2344, opened 2019-01-22
NEPT-2186: Add CKEditor plugin for link target to remove popup window and frame options

ku-enza:
### Description

Add CKEditor plugin for target link that remove  <popup window><frame> options

### Change log

- Added: JS plugin file 
- Changed: wysiwyg multisite_wysiwyg_ckeditor_plugin , multisite_wysiwyg_wysiwyg_plugin function


enriquelacoma (approved):

AlbaParrilla (approved):

Fefaine (changes requested):
I would like to reopen this ticket because i see several issues

1- the new plugin is not enabled, so that won't solve the user problem unless we explain the user they need to enable it for each profile,
2- There should be a hook_update that does that in 2.6
3- The name of the plugin is confusing, it looks like it allows to choose the options, while it only removes 2 options.  Besides it's strange to give the option to do something that should be mandatory.
+ I would like a test to be added on this, since it's our custom "feature"

I see this other suggestion implying to patch ckeditor library like it's detailed here http://handsomedogstudio.com/ckeditor-set-default-target-blank 
This has the advantage to not required any hook_update, applies to all filters and is not optional.

There is also a better option provided in the ACF config space of the profile, and a solution was developped for D8, this can be used as a basis of inspiration imo
https://www.drupal.org/project/drupal/issues/1936392

enriquelacoma:
I created a new branch from release-2..5.138 cherry picking the commits from this one (https://github.com/ec-europa/platform-dev/pull/2581)

PR #2351, opened 2019-01-30
NEPT-756: Remove the hard config that forces the settings of comments by default

AlbaParrilla:
### Description

This PR includes all the changes to remove the hard config that forces the settings of comments and set it in soft configuration. The changes included are related to several tickets, and not only NEPT-756, because each ticket asks to modify one variable and they are all going to be modified in the same PR.
The tickets included in this PR are: NEPT-755, NEPT-756, NEPT-759, NEPT-760 and NEPT-761.

### Change log

- Added:
  - functions to set the variables for comments in profiles/common/modules/custom/multisite_config/lib/Drupal/comment/Config.php
  - call the new functions to set all the variables of comments in the hook_install

- Removed: 
  - hard config of comments from all the features that include content types.


profiles/common/modules/custom/multisite_config/lib/Drupal/comment/Config.php
```diff
@@ -46,7 +58,95 @@ public function setDefaultCommentForContentType($content_type, $value) {
...
+  /**
+   * Set comments per page setting for a specific content type.
+   *
+   * @param string $content_type
+   *   Content type machine name.
+   * @param string $value
+   *   Value to be set (human readable).
+   */
+  public function setNumberCommentForContentType($content_type, $value) {
+    switch ($value) {
```
aritomelo, line 111: Where these cases come from, are they random? The ticket says to remove the default 50, but nothing regarding the options.

profiles/common/modules/features/e_library/e_library_core/e_library_core.install
```diff
@@ -115,3 +126,14 @@ function e_library_core_update_7120() {
...
   }
 }
+
+/**
+ * Set new value for e_library feature in database.
+ *
+ * Variables from comments settings must be removed from hard config.
+ * Recreate the value of the feature in database without the variable.
+ */
+function e_library_core_update_7121() {
```
aritomelo, line 143: I do understand why the hook is needed however it does not match the ticket description, do we have agreement on use it or not?

aritomelo (dismissed):

profiles/common/modules/custom/multisite_config/lib/Drupal/comment/Config.php
AlbaParrilla: These cases have been taken from the options given in the select field for the number of comments in a content type. You can check it for example in /admin/structure/types/manage/article_en.
As the given options are 10, 30, 50, 70, 90, 150, 200, 250 and 300, these are also the cases that I have added to  the function.

profiles/common/modules/features/e_library/e_library_core/e_library_core.install
AlbaParrilla: This function is needed to have the new value of the feature without reverting it. If we don't do it this way, the modified features should be reverted and some of them mustn't. It is because of these hook updates that the PR is now based in 2.6 and not in 2.5

profiles/common/modules/features/cce_basic_config/cce_basic_config.info
```diff
@@ -113,10 +113,6 @@ features[password_policy][] = Example policy
 features[taxonomy][] = classification
 features[variable][] = clean_url
 features[variable][] = comment_anonymous_article
```
Fefaine: You should remove all references to comments, we want user to decide themselves if they want comments and how they should appear, be set up

Fefaine (changes requested):

profiles/common/modules/features/cce_basic_config/cce_basic_config.info
AlbaParrilla: I just left comment_anonymous_xxx, following the instructions of this document https://webgate.ec.europa.eu/CITnet/confluence/pages/viewpage.action?pageId=582549891  which says that we must keep this variable in hard configuration.

drishu (approved):

kameleon69 (approved):

Fefaine (approved):

PR #2354, opened 2019-02-01
NEPT-948: Cleanup multisite_wysiwyg

AlbaParrilla:
### Description

The feature tries to create a wysiwyg profile that cannot be created because it uses a format that doesn't exist. 

### Change log

- Removed: profiles/common/modules/features/multisite_wysiwyg/multisite_wysiwyg.features.wysiwyg.inc


ku-enza (approved):

aritomelo (approved):
Ready, the solution is working ok.

PR #2356, opened 2019-02-05

enriquelacoma:
### Description

Remove permissions for role cem

### Change log

- Added: Add hook update to remove cem role
- Removed: hook update with permissions configuration


Fefaine (approved):
OK

PR #2357, opened 2019-02-05

enriquelacoma:
### Description

COMM_CEM LDAP group should be have the administrator role on the site

### Change log

- Changed: COMM_CEM LDAP group matching to administrator
- Removed: cem permissions and cem role


kameleon69 (commented):
You should update CEM to Administrator in the comment of ecas.inc line 398

Fefaine (approved):
OK

kameleon69 (approved):

PR #2371, opened 2019-02-12
NEPT-2100: Fix the patch to test db exists before trying to sanitize.

Fefaine:
### Description

Add a condition around sql sanitization of tables

### Change log

- Changed: Patch d.o/3017665


AlbaParrilla (approved):
Ok for  me

PR #2381, opened 2019-02-19
NEPT-2249: remove end of file tags.

Fefaine:
### Description

remove end of file tags.

### Change log

- Removed: closing php tags in phpcommon library


aritomelo (approved):
All tags removed.

PR #2414, opened 2019-03-12
NEPT-510: multisite_drupal_standard is useless.

AlbaParrilla:
### Description

The feature multisite_drupal_standard is empty and has to be removed from the platform. 
- Remove dependency to multisite_settings_core to avoid issues at uninstall (release-2.5.137)
- Uninstall it from all subsites in the fpfis-scheduler script for 2.6
- Remove the code in the next major release (release-2.6)

### Change log

- Removed: dependencies of the module to avoid issues while uninstalling.


Fefaine (approved):
OK for me, i only changed the commands, cf see ticket and the link to the "Procedure for module removals"

PR #2426, opened 2019-03-15
NEPT-2330: Administrator role is removed 

Fefaine:
### Description

Administrator role is removed when user not in CEM LDAP group

### Change log

- Changed: Conditions for removing administrator role


Fefaine:
We will include this fix in a new PR, for the moment we will revert the feature commit

PR #2436, opened 2019-03-21
NEPT-2353: Fix custom rules not applied.

drishu:
### Description

Fix custom rules not applied when alias changes, add defaut rules only if enabled.

### Change log

- Fixed: Updated node paths.


Fefaine (changes requested):
The previous alias should be cleared after the title is changed

Fefaine:
replaced with https://github.com/ec-europa/platform-dev/pull/2591

PR #2442, opened 2019-03-25

enriquelacoma:
### Description

COMM_CEM LDAP group should be have the administrator role on the site

### Change log

- Changed: COMM_CEM LDAP group matching to administrator
- Removed: cem permissions and cem role


profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -402,22 +402,20 @@ function ecas_ecas_sync_user_info($user, array $user_info, array &$edit, array $
...
   }
-
-  $cem_role = user_role_load_by_name(CCE_BASIC_CONFIG_CEM_ROLE_NAME);
+  $administrator_role = user_role_load_by_name("administrator");
   $user_roles = (isset($user->roles)) ? $user->roles : array();
   // By default, the role is not set for any user automatically.
   // Then, unset it by default and reassign it later if the user has the
   // EU login group.
-  unset($user_roles[$cem_role->rid]);
+  unset($user_roles[$administrator_role->rid]);
```
kameleon69: Admin role is removed but you should add/keep this role if the user is already grant as Administrator

kameleon69 (dismissed):

profiles/common/modules/custom/ecas/includes/ecas.inc
enriquelacoma: This has been removed

drishu (approved):

PR #2449, opened 2019-03-25

enriquelacoma:
### Description

Remove permissions for role cem

### Change log

- Removed: hook update that removed cem role


drishu (approved):

Fefaine (dismissed):
Some cem permissions are still exported, after the hook_update they will reappear if we proceed with a fr

See NEPT ticket for examples

PR #2457, opened 2019-03-28
NEPT-2293: Flush cache when content was deleted.

drishu:
### Description

Implement the varnish cache flushing at node delete time.

### Change log

- Added: hook_node_delete and hooh_path_delete for nodes with varnish purge rules.
- Changed: TokenContext to handle last deleted node id in token replacement.

### Commands

drush cc all


AlbaParrilla (approved):
Ok for me 👍

Fefaine:
Ok for me it's good, whatever the order as long as all paths are sent, that's perfect !

Fefaine (approved):
OK

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+  // No rule defined for the node: Ignore the rest of the process.
+  if (!nexteuropa_varnish_node_has_rules($node)) {
+    return;
+  }
+
+  $paths = array();
+
+  $entity_uri = entity_uri('node', $node);
+
+  if ($entity_uri) {
```
drupol: Just for the sake of maintainability in the future it would be nice to avoid this kind of "ghosts" `if` conditions.
I don't say it's wrong, but as PHP is not strongly typed (yet?) and even less Drupal, it would be easier for everyone to have a better condition there.

From the API documentation, [entity_uri()](https://api.drupal.org/api/drupal/includes%21common.inc/function/entity_uri/7.x) returns:
_An array containing the 'path' and 'options' keys used to build the URI of the entity, and matching the signature of url() or NULL if the entity has no URI of its own._

Maybe the proper condition here would have been:

`if (null !== $entity_uri)` or `if (is_array($entity_uri))`, with a preference for the first one, which is faster.

drupol (changes requested):

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+    $url_locale = language_negotiation_get('language', 'locale-url');
+    $url_suffix = language_negotiation_get('language', 'nexteuropa_multilingual_url_suffix');
+    // Add logic to send a language prefix/suffix regex for non aliased paths.
+    if ($url_locale) {
```
drupol, line 245: Same here...

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+      $lgregex = '[a-z]{2}\/';
+    }
+    elseif ($url_suffix) {
```
drupol, line 248: Same here...

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+      $lgregex = '_[a-z]{2}';
+    }
+    else {
+      $lgregex = '';
```
drupol: To decrease the cyclomatic complexity, the `else` condition could be removed here.
```php
$lgrexex = '';

if ($url_locale) {
 // stuff
} elseif ($url_suffix) {
 // stuff
}
```

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+    $paths['escaped_alias'] = preg_quote($alias, '/');
+    // Add regex prefix or suffix if needed to alias.
+    // Example content\/my\-page_[a-z]{2}.
+    $paths['alias_prefix'] = $url_suffix ? $paths['escaped_alias'] . $lgregex : $lgregex . $paths['escaped_alias'];
+  }
+
+  // Add custom purge rules.
+  $purge_rules = nexteuropa_varnish_get_node_purge_rules($node);
+  foreach ($purge_rules as $purge_rule) {
+    if ($purge_rule->type() == PurgeRuleType::PATHS) {
```
drupol, line 270: Use `===`

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+      $purge_paths = $purge_rule->paths();
+      $paths = array_merge($paths, $purge_paths);
```
drupol: There is no need to have `$purge_paths` as a variable.

Use: `array_merge($paths, $purge_rule->paths());`

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+    }
+  }
+
+  if (!empty($paths)) {
+    // Removing duplicates and all entries of the array equal to FALSE.
```
drupol: This comment makes no sense.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -223,6 +223,110 @@ function nexteuropa_varnish_node_presave($node) {
...
+ */
+function nexteuropa_varnish_path_delete($path) {
+  if (empty($path)) {
+    return;
+  }
+
+  $node = menu_get_object('node', 1, $path['source']);
+
+  // Its a node path.
+  if (!empty($node)) {
```
drupol: Maybe an early return here would be better, just like you did with `$path`.

```php
if (empty($node)) {
  return;
}
```

It lower the complexity of the code, it's consistent with the rest, and the main code doesn't gets wrapped by a condition.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -546,63 +650,91 @@ function _nexteuropa_varnish_get_node_paths($node) {
...
     if (module_exists('entity_translation') && entity_translation_enabled('node', $node)) {
       $handler = entity_translation_get_handler('node', $node);
       $translations = $handler->getTranslations();
       $lang_codes = array_keys($translations->data);
       $all_languages = entity_translation_languages('node', $node);
-      foreach ($lang_codes as $lang_code) {
-        if (isset($all_languages[$lang_code])) {
-          $alias = drupal_get_path_alias($original_path, $lang_code);
+      // Small workaround for undefined language.
+      if (count($lang_codes) == '1' && 'und' == $lang_codes[0]) {
```
drupol: use `===`

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -546,63 +650,91 @@ function _nexteuropa_varnish_get_node_paths($node) {
...
+        // Push all the languages.
+        $lang_codes = array_keys($all_languages);
+      }
+      foreach ($lang_codes as $code) {
+        if (isset($all_languages[$code])) {
```
drupol: You could also use an early return here:

```php
if (!isset($all_languages[$code])) {
  continue;
}
```

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -546,63 +650,91 @@ function _nexteuropa_varnish_get_node_paths($node) {
...
+          $alias = drupal_get_path_alias($entity_uri['path'], $code);
           $original_alias = $alias;
-          // Do not use altered "options" for url_outbound.
-          $altered_options = $options + array('language' => $all_languages[$lang_code]) + array('external' => FALSE);
-          // Add the suffix/prefix depending negociation setup.
-          drupal_alter('url_outbound', $alias, $altered_options, $original_path);
-          $paths[$lang_code] = preg_quote($alias, '/');
+          $lang = $all_languages[$code];
```
drupol: Where `$lang` is used elsewhere ? Why do we need a variable for it ?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
```diff
@@ -546,63 +650,91 @@ function _nexteuropa_varnish_get_node_paths($node) {
...
       $original_alias = $alias;
 
       // Language is undefined, the content will show on all languages.
       if ('und' == $lang_code) {
-        foreach ($languages as $code => $data) {
-          $altered_options = $options + array('language' => $languages[$code]) + array('external' => FALSE);
-          // Add the suffix/prefix depending negociation setup.
-          drupal_alter('url_outbound', $alias, $altered_options, $original_path);
-          $paths[$code] = preg_quote($alias, '/');
+        foreach ($all_languages as $code => $data) {
```
drupol: `$code` and `$data` are not very meaningful. It would be nice to use better variable name.

drupol (changes requested):
I did some remarks. I think it would be nice to implement them.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: After re-reading the comment today, I understood it!
Maybe such comment is not necessary. Everyone is supposed to know what `array_filter()` and `array_unique()` is doing, they are pure PHP functions.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: Fixed.

drupol (dismissed):
Thanks for fixing a lot of comments, it's much better.

Now, could you either comment on the remaining remarks or fix them in a next commit ?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: `if (count($lang_codes) === 1 && LANGUAGE_NONE === $lang_codes[0])`

Count returns a int, so, don't worry for this one. and I think (to be verified) that 'und' could be replaced with a constant ([LANGUAGE_NONE](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/constant/LANGUAGE_NONE/7.x)), so this is a string and triple equals can be used without any troubles.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drishu: language_negotiation_get return an ID or FALSE, if evaluates for TRUE or FALSE so it would be redundant.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drishu: language_negotiation_get return an ID or FALSE, if evaluates for TRUE or FALSE so it would be redundant.

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drishu: seems to risky to type compare on something entity_load returns

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drishu: Makes sense for count, the other is again to risky

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: If $url_locale ! == false then ?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: `elseif $url_suffix ! == false` then ?

profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
drupol: In this case we can use === because we are comparing strings no ?

PR #2464, opened 2019-03-29
NEPT-2391: Provide default destination for when session is closed.

Fefaine:
### Description

Default destination of ecas login set to frontpage

### Change log

- Added: Default value for destination variable at login


drishu (approved):

tomv-be (approved):
Code is ok, change works as expected.

PR #2469, opened 2019-04-02
NEPT-2379: Break and li tags in a single cell in excell export.

drishu:
### Description

XLS Export renders correctly, break and li tags represent in a single cell with inline wrap instead of multiple cells.

### Change log

- Added: Patch https://www.drupal.org/files/issues/2019-04-02/break-and-li-tags-single-cell-2400129-8.patch.


Fefaine:
Rejected, see nept ticket

PR #2474, opened 2019-04-03
NEPT-2399: Not removing slash if redirect target is external.

drishu:
### Description

Target URL was modified when implementing redirection using Redirect module, so we are not removing slash if redirected target is external.

### Change log

- Added: Patch target-modified-implementing-redirection.patch.
- Removed: https://www.drupal.org/files/issues/2018-06-24/redirect-increase-size-fields-to-900-2057615-35.patch.
- Changed: Redirect module version to commit d5c9521.


AlbaParrilla (approved):
Ok for me

PR #2475, opened 2019-04-03
NEPT-2318: Upgrade views to 3.21 for release-2.5.

AlbaParrilla:
### Description

Upgrade views to the last version. It includes 3 security patches so the upgrade (or a security patch) must be included also in 2.4 and 2.5.

### Change log

- Changed: resources/multisite_drupal_standard.make


resources/patches/issues-ajax-exposed-filters-blocks.patch
```diff
@@ -1,30 +0,0 @@
-diff --git a/js/ajax_view.js b/js/ajax_view.js
```
kameleon69, line 1: This patch is still needed?

kameleon69 (changes requested):

PR #2484, opened 2019-04-09
NEPT-2384: Enable the contact module before configuring the menu

enriquelacoma:
### Description

Enable the module contact before configuring the menu.

### Change log

- Changed: The process to enable the dependency of contact module.


tomv-be (dismissed):
Code looks fine. Ran into an error (unrelated, probably), but the code seems to work as expected.

tomv-be (approved):
Feature enables without any error now. Works as expected.

AlbaParrilla (approved):

PR #2514, opened 2019-04-26
NEPT-1606: Implement drone on 5.6

Fefaine:
### Description

Implement drone on 5.6

### Change log

- Changed: Several tests and behat extensions


tests/behat.yml.dist
```diff
@@ -65,7 +65,7 @@ default:
       application:
         base_url: '{{ behat.base_url }}'
         endpoint: '/poetry/notification'
-    Drupal\MinkExtension:
+    Behat\MinkExtension:
```
drupol: I think this needs to stay `Drupal\MinkExtension` now that we are using the beta channel.

Fefaine: YES sorry i was rolling back all the changes to the stable (green) state, it's now fixed!
🎉

drupol:
Great news 💖

PR #2516, opened 2019-04-29
NEPT-536: Merge multisite_og_button into multisite_create_button

AlbaParrilla:
### Description

Merge both custom modules. All these modules do is to create the button "create content" in standard profile and communities (one module for each profile).
 
### Change log

- Added: merged content from multisite_og_button into multisite_create_button.
- Removed: multisite_og_button module


Fefaine:
replaced with https://github.com/ec-europa/platform-dev/pull/2602

PR #2526, opened 2019-05-06
NEPT-2439: Remove nexteuropa_integration.

AlbaParrilla:
### Description

Remove nexteuropa_integration feature

### Change log

- Removed: nexteuropa_integration feature


drishu (dismissed):

tomv-be (changes requested):
Looks good, but there are still symlinks left in the multisite_drupal_standard and multisite_drupal_communities profiles, I guess they should be removed?

PR #2528, opened 2019-05-07
NEPT-1017: Remove ec_resp from default values

AlbaParrilla:
### Description

Remove ec_resp from default values. 
There still is a reference to ec-resp without taking into account of ec_europa: 
https://github.com/ec-europa/platform-dev/blob/223bd1dddd5bc4d29dd3ce5b7a7f25d7e6d2e14a/profiles/common/modules/features/multisite_custom_error/multisite_custom_error.install#L19
The other references are for treating a case specific for ec_resp.

### Change log

- Removed: unuseful variable from multisite_custom_error feature, which was set to 'ec_resp' by default.


tomv-be (approved):
Ok for me.

drishu:
Reopening due to automatic close.

PR #2542, opened 2019-05-14
NEPT-2079: Fix issues with enabling simpletests

enriquelacoma:
### Description

Update test implementation

### Change log

- Changed: Changed the classes and method used on the test


tomv-be (approved):
Fixes the error. Ok for me.

PR #2564, opened 2019-05-20
NEPT-1343: Fix bug, theme & add the multisite_og_button block to the relevant region

enriquelacoma:
### Description

Missing content selector on communities on eu_europa when the install theme is ec_resp

### Change log

- Added: cce_basic_config_themes_enabled() hook to configure blocks for ec_europa
- Removed: remove enabling of ec_europa blocks on install


enriquelacoma:
Will be implemented for 2.6

PR #2583, opened 2019-05-28
NEPT-2186: WYSIWYG - Remove unauthorised target option JAVASCRIPT under popup window "LINK"

enriquelacoma:
### Description

Remove popup and iframe options from the target options of the Link Ckeditor plugin.
Update: the iframe option should not be removed, only the popup.

### Change log

- Added: File link_alter.js to remove options from the target select.


AlbaParrilla (approved):

AlbaParrilla (approved):
Ok for me. Option iframe is again available, while pop-up option has been removed.

Fefaine (approved):

PR #2607, opened 2019-06-20
NEPT-2527: CLONE - COMMISSION - Site under maintenance. The website encountered an unexpected error.	

AlbaParrilla:
### Description

Before the last deployment there was a problem with some  sites using media AV portal. A maintenance page appeared because some videos couldn't be shown. A patch to fix that exception when a video cannot be found has been added.

### Change log

- Added: patch for media AV portal module


drupol:
The `try/catch` block should only contain the part of the code that might return an exception, something [like this](https://paste.ofcode.org/D4gKrLpzSNh8Lb3bmRatNv) might be more appropriate.

In order to have a better readability and maintainability, I'd move the return out of the `try/catch` block as well.

Where is defined the hook_theme: `media_avportal_broken` ? I checked the media_avportal module, but it's not in there.
If it's somewhere else, then its name is wrong. The name of the container module should be the prefix of the hook_theme.

Fefaine (changes requested):
This patch seems too complicated, and as @drupol  says the tpl that is called does not exist.
Furthermore, i don't think we should call a different template, that webservice is only used here to check the language and there is a default ('en') for the first occurence of the call, so i would just do this https://paste.ofcode.org/yAUBLPWP3nzvNf3ggJvhut
For the second call, i would use the default thumbnail provided by media module if the thumbnail cannot be returned

resources/patches/media_avportal-improve-exception-management.patch
```diff
@@ -0,0 +1,35 @@
...
+-    $languages = $wrapper->getAvailableVideoLanguage();
++    try {
++      $languages = $wrapper->getAvailableVideoLanguage();
++    }
++    catch (Exception  $e) {
++      watchdog_exception('media_avportal', $e);
++    }
+     $video_language = 'en';
+-    if (in_array($language->language, $languages)) {
++    if (isset($languages) && in_array($language->language, $languages)) {
```
drupol:
```suggestion
+    if (is_array($languages) && in_array($language->language, $languages)) {
```

resources/patches/media_avportal-improve-exception-management.patch
```diff
@@ -0,0 +1,35 @@
...
+       $video_language = $language->language;
+     }
+     $options['lg'] = $video_language;
+@@ -113,7 +118,12 @@ function media_avportal_file_formatter_image_view($file, $display) {
+   $wrapper = file_stream_wrapper_get_instance_by_uri($file->uri);
+   $image_style = $display['settings']['image_style'];
+   $valid_image_styles = image_style_options(FALSE);
+-  $path = $wrapper->getLocalThumbnailPath();
++  try {
++    $path = $wrapper->getLocalThumbnailPath();
```
drupol: :heart:

drupol (changes requested):

resources/patches/media_avportal-improve-exception-management.patch
drupol: So much better than the previous patch!

resources/patches/media_avportal-improve-exception-management.patch
```diff
@@ -0,0 +1,35 @@
...
++  }
++  catch (Exception  $e) {
++    watchdog_exception('media_avportal', $e);
```
Fefaine: That's nice to have an exception, but here you don't have a $path variable
That's why i was suggesting to use the default media thumbnail for this.
It should be located at /sites/default/files/media-icons/default/video-x-generic.png and can be accessed using :
public://media-icons

Otherwise we'd still have a fatal error if ws is not available

AlbaParrilla: I've added the default media thumbnail and I've removed the exception. There is already an exception inside getLocalThumbnailPath() in which the default thumbnail would be used in case the thumbnail is not available.

Fefaine: Ok good catch, so that's fine ! I'll make a new test now !

resources/patches/media_avportal-improve-exception-management.patch
```diff
@@ -0,0 +1,33 @@
...
+-    $languages = $wrapper->getAvailableVideoLanguage();
++    try {
++      $languages = $wrapper->getAvailableVideoLanguage();
++    }
++    catch (Exception  $e) {
++      watchdog_exception('media_avportal', $e);
++    }
+     $video_language = 'en';
+-    if (in_array($language->language, $languages)) {
++    if (is_array($languages) && in_array($language->language, $languages)) {
```
Fefaine: This throws an error when the ws is unavailable. The variable $languages is not set.

Fefaine (approved):

PR #2614, opened 2019-06-24
NEPT-2174: Implement security patch on password policy.

drishu:
### Description

Implement security patch on password policy forward-port-of-sa-contrib-2018-077.patch.

### Change log

- Added: Patch forward-port-of-sa-contrib-2018-077.patch.


drupol (approved):
:+1:

PR #2617, opened 2019-06-25
NEPT-2529: Show EU Login logout menu if usernames are different

tomv-be:
### Description

Show the EU Login logout link when logged in with EU Login, even if the EU Login username does not match the Drupal one.

### Change log

- Changed: Modified normal_menu_logout_check() and ecas_menu_logout_check() to not require that the ecas and drupal usernames match

### Commands

drush cc all


Fefaine (approved):
Ok for me, tested with external accounts and internal

PR #2620, opened 2019-06-26
NEPT-2503: Update ApacheSolr

tomv-be:
### Description

- Update Apache Solr Search to version 7.x-1.11
- Modified patch to apply correctly

### Change log

- Added: new local patch, modified from the drupal.org version
- Changed: updated .make file to use the new module version and point to the local patch

### Commands

drush cc all


AlbaParrilla (approved):
Ok for me

drishu (approved):

PR #2644, opened 2019-07-16
NEPT-2439: Remove nexteuropa_integration from 2.6 and all the modules that are useless without it.

AlbaParrilla:
### Description

Module nexteuropa-integration was already removed. The ticket has been reopened in order to remove the modules that are useless now that we don't have nexteuropa_integration module anymore.
These useless modules are:
- integration
- integration_couchdb
- couchdb
- integration_test

### Change log

- Removed: the modules named in the description


Fefaine (approved):
OK

drishu (approved):

PR #2647, opened 2019-07-18
NEPT-2576: Moved nexteuropa_varnish to separate repository.

drishu:
### Description

Moved nexteuropa_varnish to separate repository.

### Change log

- Added: ec-europa/digit-ne-varnish repository to make file
- Removed: nexteuropa_varnish module from profiles folder

### Commands

drush rr


kameleon69:
There is a conflict

resources/multisite_drupal_standard.make
```diff
@@ -1194,3 +1194,9 @@ projects[nexteuropa_poetry][type] = module
 projects[nexteuropa_poetry][download][type] = git
 projects[nexteuropa_poetry][download][url] = https://github.com/ec-europa/nexteuropa_poetry.git
 projects[nexteuropa_poetry][download][tag] = 0.1.0
+
+projects[nexteuropa_varnish][subdir] = "contrib"
```
kameleon69, line 1198: It should be "custom" and not contrib, no ?

kameleon69 (changes requested):

resources/multisite_drupal_standard.make
drishu: That's debatable, nexteuropa_poetry is also in contrib, for a reason I imagine. Also, if I move it into custom it will be picked up by git.

PR #2648, opened 2019-07-19
NEPT-2588: Prevent any process if the purge is disabled.

Fefaine:
### Description

Prevent any process if the purge is disabled.

### Change log

- Added: Additional check before sending purge request. This is to prevent subsites code sending purge request using the callback function to go through when purge is disabled.


kameleon69 (approved):

AlbaParrilla (approved):
Ok for me

PR #2651, opened 2019-07-22
NEPT-2590: Upgrading to 7.2

Fefaine:
### Description

Upgrading to 7.2

### Change log

- Changed: drone php version


Fefaine:
Drone is freezing again, i take the risk to merge it because i'm really tired of waiting for buggy systems to do their jobs

PR #2654, opened 2019-07-23
Release 2.5.138.3

Fefaine:
### Description

Prevent any purge request being sent if blocking variable was set

### Change log

- Added: Extra check in private function


iulianweb (commented):
All fine, please proceed.

PR #2664, opened 2019-07-29
Nept 2524 : Set CKEditor library to the right version.

kameleon69:
## NEPT-[Insert ticket number here]
Nept 2524
### Description
Set CKEditor library to the right version.
[Insert description here]

### Change log

- Changed: version of CKEditor in wysiwyg profile.

### Commands

drush updatedb


kameleon69:
bad reference

PR #2671, opened 2019-08-12
NEPT-2178: Translation workflow may not ensure that text removed from original is also removed from translations

Fefaine:
### Description

Translation workflow may not ensure that text removed from original is also removed from translations

### Change log

- Added: patch on d.o and patch in tmgmt_workbench custom module

### Commands

```drush cc all ```


AlbaParrilla (approved):
Ok for me 👍

PR #2677, opened 2019-08-19
NEPT-2201: Hide username from RSS feed when content type is set to not display author.

Fefaine:
### Description

Hide username from RSS feed when content type is set to not display author.

### Change log

- Added: Core patch


kameleon69 (approved):

AlbaParrilla (approved):
Ok for me

PR #2683, opened 2019-08-23
NEPT-2641: Fix func_get_args() no longer report the original value.

drishu:
### Description

Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as passed to a parameter

### Change log

- Added: Patch views-php7-3076826-2.patch


AlbaParrilla (approved):

Fefaine (approved):
OK

Fefaine (approved):
OK for me

PR #2685, opened 2019-08-25
NEPT-2643: Fix func_get_args() no longer report the original value.

drishu:
### Description

Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as passed to a parameter

### Change log

- Added: Exclude rule in phpcs-compact.xml
- Changed: xml_field version to 2.3


Fefaine (approved):
OK

AlbaParrilla (approved):

PR #2689, opened 2019-08-27
NEPT-2652: Entity Translation patch replacement

AlbaParrilla:
### Description

This patch that we are using 
https://www.drupal.org/node/2741407

Will be replaced by this config feature
https://www.drupal.org/project/entity_translation/issues/3010146

This task is to replace the first patch with the second patch

### Change log

- Changed: replaced patch for entity_translation in multisite_drupal_standard.make


drishu (approved):

Fefaine (approved):
OK

PR #2692, opened 2019-08-28
NEPT-2651: Add a hook that disables video if no users have added videos.

Fefaine:
### Description

Add a hook that disables video if no users have added videos.

### Change log

- Added: cce_basic_config_update_7234
- Removed: Hard config for video fields and dependency

### Commands

```drush updb```


tomv-be (approved):

kameleon69 (changes requested):
One case is missing see Jira comment

kameleon69 (approved):

PR #2693, opened 2019-08-28
NEPT-520: Prevent EU Login to create 2 users with the same email.

drishu:
### Description

Prevent EU Login to create 2 users with the same email, Do not allow user that have other roles than "Authenticated" to login using social network, but force them to use EU Login form.

### Change log

- Added:
  - ecas.variable.inc
  - theme/templates/ecas-profile-warning-page.tpl.php 
- Changed: Refactored code
- Security:
  - Prevent duplicate email
  - Prevent social login if user is other then authenticated

### DIFF
--- a/profiles/common/modules/custom/ecas/includes/ecas.inc
+++ b/profiles/common/modules/custom/ecas/includes/ecas.inc
@@ -6,11 +6,6 @@
  */
 
 /**
- * ECAS group for CEM team users.
- */
-define('ECAS_CEM_ECAS_GROUP', 'COMM_CEM');
-
-/**
  * Checks if the given user is an ECAS user.
  *
  * @param object $user
@@ -27,6 +22,7 @@ function _is_ecas_user($user) {
       }
     }
   }
+
   return FALSE;
 }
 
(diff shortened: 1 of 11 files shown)


profiles/common/modules/custom/ecas/includes/ecas.inc
```diff
@@ -441,133 +506,170 @@ function ecas_login_check() {
...
-    // When returning with same ECAS login as before closing Drupal session.
-    else {
-      if (
-        isset($account->data['ecas_login_date']) &&
-        $account->data['ecas_login_date'] == $user_info['cas:loginDate']
-      ) {
-
-        $msg = 'Rejected ECAS reused login: %n (ECAS)';
-        watchdog('user', $msg, array('%n' => $account->name), WATCHDOG_NOTICE);
+  if ($ecas_social_login == 'v' && (!user_has_role($authenticated_role->rid, $account) || count($account->roles) > 1)) {
```
tomv-be, line 622: This condition will prevent users from signing up with a Social Media account. As the account has not yet been created, it will not have the Authenticated User role and will get the error message.

PR #2704, opened 2019-09-02
NEPT-2655: views_data_export is not compatible with PHP 7.2

AlbaParrilla:
### Description

views_data_export is not compatible with PHP 7.2

### Change log

- Added: patch to fix compatibility issue with php 7.2


drishu (approved):

Fefaine:
drone is red, but what can we expect anymore...

PR #2707, opened 2019-09-04
NEPT-2657: simplenews is not compatible with php 7.2

AlbaParrilla:
### Description

Add patch to fix compatibility issues with php 7.2

### Change log

- Added: patch in resources/multisite_drupal_standard.make


tomv-be (dismissed):

Fefaine (approved):
OK

PR #2719, opened 2019-09-10
NEPT-2668: Make Language selector compatible with content translation.

drishu:
### Description

Make Language selector compatible with content translation, language selector did not work when multilingual support was enabled with node translation.

### Change log

- Added: Patch nept-2668-language-switcher.patch
- Changed: ec_europa_preprocess_lang_select_site()


AlbaParrilla (approved):

Fefaine (changes requested):
We should not use a dev branch here

PR #2739, opened 2019-10-10
NEPT-536: Merge multisite_og_button into multisite_create_button

AlbaParrilla:
### Description

Merge both custom modules. All these modules do is to create the button "create content" in standard profile and communities (one module for each profile).
This PR adds the code to fix NEPT-536 and NEPT-1343

### Change log

- Changed: merged content from multisite_og_button into multisite_create_button.
- Removed: multisite_og_button module


profiles/common/modules/custom/multisite_create_button/multisite_create_button.module
```diff
@@ -27,6 +31,11 @@ function multisite_create_button_block_view($delta = '') {
       $block['subject'] = t('Create content');
       $block['content'] = _multisite_create_button_contextual_links();
       break;
+
+    case 'og-contextual-links':
+      $block['subject'] = t('Create content');
+      $block['content'] = _multisite_og_button_contextual_links();
```
drishu: this should be renamed into _multisite_create_button_contextual_links()

drishu (changes requested):
Also symlinks have to be removed

profiles/common/modules/custom/multisite_create_button/multisite_create_button.module
AlbaParrilla: _multisite_create_button_contextual_links is the one used for the previous "create_button" and this function is different, so I renamed it as _multisite_create_button_og_contextual_links()

tomv-be (dismissed):

PR #2742, opened 2019-10-10
NEPT-2667: Implementation of global banner in 2.6

Fefaine:
### Description

Implementation of global banner

### Change log

- Changed:

### Commands

[Insert commands here]

### Checklist


Fefaine:
Won't do, see latest comment in https://webgate.ec.europa.eu/CITnet/jira/browse/NEPT-2667

PR #2770, opened 2019-10-31
NEPT-1587: Removed multisite_metatags, symlink, config.

drishu:
### Description

Remove multisite_metatags from the core platform

### Change log

- Removed: module, symlinks, configuration

### Commands

[Insert commands here]

### Checklist


julien-:

If you think it should be reviewed, ask it.

PR #2773, opened 2019-11-06
NEPT-2499: Show values for other languages than default.

Fefaine:
### Description

Fix content types with another language than English don't show their values.

### Change log

- Added: Patch entity_translation-content_not_in_current_language-2877103-17.patch

### Checklist


Fefaine:
The patch fixes the issue, we will follow up on the actions of the maintainer

PR #2774, opened 2019-11-06
NEPT-2731: patch Maxlength

tomv-be:
### Description

Add a patch for Maxlength

### Change log

- Added: patch maxlength_js_label_translatable-3001622-2.patch

### Checklist


Fefaine (approved):
OK

Fefaine:
I am closing this PR, the fix should be done on the translation feed that should occur on triggering the page translation.
Also it could go only in 2.6 it's not needed in 2.5 by anyone

PR #2776, opened 2019-11-11
NEPT-2737: Fix tracking changes regression.

drishu:
### Description

Instead of freezing current format if there are tracked changes, we should block those formats that are not "enabled" there.

### Change log

- Changed: Text profile disable behaviour.
- Fixed: Behat test for tracked changes.

### Checklist


Fefaine (approved):
Perfecto, well done :)

tomv-be (approved):

PR #2786, opened 2019-11-21
NEPT-2745: Create a Paste from Word filter

drishu:
### Description

Implements a filer activated with paste from word, which filters at next paste.

### Change log

- Added: js file with drupal behaviour

### Commands

drush cc css-js

### Checklist


Fefaine (approved):
OK

PR #2789, opened 2019-11-29
NEPT-2730: Remove unecessary translator of type 'DGT Connector'

drishu:
### Description

Permit only user 1 to delete translators

### Change log

- Changed: function _tmgmt_poetry_translator_access().

### Checklist


Fefaine (approved):
OK

PR #2835, opened 2020-01-22
NEPT-2774: Next Europa varnish generating php error.

mehdi0712:
### Description

Notice: Undefined index: prefix in _nexteuropa_varnish_build_path() (line 781 of /ec/prod/app/webroot/drupalm_home/reference-sources/multisite/multisite_master_test.2.5/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module).

The issue is that on nexteuropa varnish :
     We manage all installed languages:
           https://github.com/ec-europa/platform-dev/blob/release-2.5/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module#L706

    And supposed to add a prefix on each of them:
          https://github.com/ec-europa/platform-dev/blob/release-2.5/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module#L773
           
    But drupal core adds a prefix on enabled languages only :
      includes/locale.inc line 419

If the need of manage disabled languages is confirmed I can make another PR to allow it.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Checklist

### DIFF
--- a/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
+++ b/profiles/common/modules/custom/nexteuropa_varnish/nexteuropa_varnish.module
@@ -703,7 +703,7 @@ function _nexteuropa_varnish_get_node_paths($node) {
     else {
       // Content translation.
       $lang_code = entity_language('node', $node);
-      $all_languages = language_list();
+      $all_languages = language_list('enabled');
       $alias = drupal_get_path_alias($entity_uri['path'], $lang_code);
       $original_alias = $alias;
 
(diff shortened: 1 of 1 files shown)


Fefaine (commented):
Why not using this one https://api.drupal.org/api/drupal/modules%21locale%21locale.module/function/locale_language_list/7.x 
Since (thank god) no site disabled  locale :p

drishu:
dono how I missed this first, but using locale_language_list is nok because it only return 1 field from the language object whilst drupal_alter('url_outbound') expects a full object,.. so this should stay as is.

Fefaine:
🤦‍♀

Fefaine (approved):
Tested ok

tomv-be (approved):

PR #2851, opened 2020-02-09
NEPT-2782: Fix print button missing and pdf tool not selected.

drishu:
### Description

Enabling print_ui as the print buttons block is defined there, also redefined pdf lib path.

### Change log

- Added:
   - print_ui to cce_basic_config dependencies
   - update hook to set print pdf tool path option

### Checklist


Fefaine (changes requested):
The hook install and the hook update do not have the same path, that does not seem correct to me https://github.com/ec-europa/platform-dev/pull/2851/files#diff-ef2b9c8b246e6fcf1b918b6ec123c008L1228

Fefaine (dismissed):

Fefaine (approved):
Ok now

PR #2866, opened 2020-02-24
NEPT-2791: Fix 'json_settings' property_type for usage Entity API.

julien-:
### Description

I would like use Entity API `entity_metadata_wrapper()` function in my hook_node_preprocess but my field of type "json_settings" is not listed because the "property_type" is missing.

See : https://drupal.stackexchange.com/questions/162396/cant-access-custom-field-through-entity-metadata-wrapper

### Change log

- Added:
"property_type" for the custom field type "json_settings".


Fefaine (approved):
Ok for me

drishu (approved):

PR #2880, opened 2020-03-06
NEPT-2709: Fix undefined function current_path in registry rebuild

drishu:
### Description

Fix undefined function current_path in registry rebuild.

### Change log

- Changed: cce_basic_config_query_alter to check function_exists

### Checklist


Fefaine (changes requested):
@drishu  isn't it more efficient to return if the function does not exist or uid=1;  and only check the path afterwards.
With this code, if the function does not exist, we check if empty string equals 'admin/people'

I would rather do

if ($user->uid == 1 || !function_exists('current_path')) {
    return;
  }

Fefaine (approved):
Ok for me

PR #2881, opened 2020-03-09
NEPT-2807: Fix filesize() fail.

drishu:
### Description

Core patch, fix filesize() fail.

### Change log

- Added: core-remote_file_save-3087532-3.patch

### Checklist


Fefaine:
The patch seems is not to be added with the details of info we have today. See nept ticket

PR #2883, opened 2020-03-10
NEPT-2797: Update Atomium

tomv-be:
### Description

Atomium release 2.31 resulted in several warnings and notices.
Switched to specific revision and included a patch to prevent notices.

### Change log

- Added: Added a patch to resolve notices
- Changed: Updated Atomium to specific revision

### Checklist


Fefaine (changes requested):
The atomium patch should be proposed on drupal org

PR #2897, opened 2020-04-23
NEPT-2813: Fix wrong function ecas_sync_drupal_user_on_ecas_info call.

dxvargas:
### Description

Fix ecas function calling from `ecas_sync_drupal_user_on_ecas_info()`, to correct name `ecas_sync_drupal_user_with_ecas_info()`.

### Change log

- Fixed: ecas function call
### Checklist

### DIFF
--- a/profiles/common/modules/custom/ecas/includes/ecas.inc
+++ b/profiles/common/modules/custom/ecas/includes/ecas.inc
@@ -375,11 +375,11 @@ function ecas_sync_profile_core(&$user, array $user_info, array &$edit = array()
  *   info_ecas_update() hook.
  *
  * @deprecated
- *   Use the combination of ecas_sync_drupal_user_on_ecas_info() and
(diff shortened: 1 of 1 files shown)


drishu (approved):
I was able to reproduce the issue with implementing hook_ecas_sync_user_info and calling ecas_sync_user_info from there, then you get Maximum function nesting level of '256' reached, but it all depends on how that custom code is impemented. Not our concern since ecas_sync_user_info is marked as deprecated.
The undefined function is replaced indeed, so its ok for me.

EDIT: implementation is https://github.com/ec-europa/eu-aid-reference/blob/61c8926b861d78fe44b3094ccaa3bff61c83b977/lib/modules/features/euaid_vacancy_assignment/forms/euaid_vacancy_assignment.deploy_volunteer.form.inc#L103 which is safe from the max function limit.

PR #2904, opened 2020-05-18
NEPT-2833: Cannot add languages to ongoing translation request (on specific content types)

enriquelacoma:
### Description

Change the language check to enable the "Request additional languages" option

### Change log

- Changed: _tmgmt_poetry_get_translatable_languages() 

### Checklist


julien-:

If you think it should be reviewed, ask it.

PR #2916, opened 2020-06-24
NEPT-2850: SA-CONTRIB-2020-025

tomv-be:
### Description

Patch to remedy SA-CONTRIB-2020-025

### Change log

- Security: Patch for i18n to prevent XSS

### Checklist


Fefaine (approved):
checkplain added

PR #2929, opened 2020-09-25
NEPT-2872: Restore file that was erroneously removed in NEPT-2823.

Fefaine:
### Description

[Insert description here]

### Change log

- Added: Restore file that was erroneously removed in NEPT-2823.

### Checklist


Fefaine:
Won't do, will be replaced with NEPT-2868

PR #2932, opened 2020-09-29
NEPT-2876: Fix the condition in the drupal access module.

Fefaine:
### Description

Fix the condition in the drupal access module.

### Change log

- Fixed: Incorrect condition in the hook_grant

### Commands

[Insert commands here]

### Checklist


zarabatana (approved):

Fefaine:
Also checked with Lux communities team

PR #2936, opened 2020-10-02
Nept 2876

Fefaine:
### Description

Fix the condition in the drupal access module.

### Change log

- Fixed: Fixed: Fix the condition in the drupal access module.

### Checklist


Fefaine:
outdated

PR #2940, opened 2020-10-23
NEPCB-78: Feature generation according Toolkit PHPCS rules.

julien-:
### Description

Add automatically the 'php' property to the .info file of generated feature, as required by PHPCS.

### Change log

- Added:
 Module 'nexteuropa_formatters_features'

### Commands

drush fe your_feature_name -y


Fefaine:
The rule will be removed.
It is impossible to add this inside the platform as the only feature that is enabled on all the sites (cce_basic_config) is never reverted.
Thank you for the PR though, we really appreciate your efforts to improve.

PR #2949, opened 2020-11-23
NEPT-2719: remove dependency from community profile.

Fefaine:
### Description

remove dependency from community profile.

### Change log

- Changed: removed dependency in the communities.info file

### Commands

[Insert commands here]

### Checklist


Fefaine:
won't do

PR #2960, opened 2021-01-19
NEPT-2895: Do not switch on the 2AF enforcement by default

Fefaine:
### Description

If the "ecas_whitelisted_user_roles" is not present, do not enforce the 2AF

### Change log

- Changed:  2AF is not enforced by default

### Checklist


Fefaine:
reviewed by Mehdi and Enrique, we will test fully in acceptance 2.6.12

PR #2967, opened 2021-02-21
NEPT-2879: Remove the problematic constant CCE_BASIC_CONFIG_CEM_ROLE_NAME

gillesdeudon:
### Description

The goal of the PR is to remove the constant that created an undeclared dependency to cce_basic_config module. 
It is also the opportunity to normalise the use of constants and string value to avoid constants that only use once and only keep constants when string values are used at different location.

### Change log

- Changed: "profiles/common/modules/custom/ecas/includes/ecas.constants.inc", remove the "ECAS_CEM_ECAS_GROUP" constant declaration.
- Changed: "profiles/common/modules/custom/ecas/includes/ecas.inc", remove the call to the constants "CCE_BASIC_CONFIG_CEM_ROLE_NAME" and "ECAS_CEM_ECAS_GROUP".

### Commands

No command execution required

### Checklist

### DIFF
--- a/profiles/common/modules/custom/ecas/includes/ecas.constants.inc
+++ b/profiles/common/modules/custom/ecas/includes/ecas.constants.inc
@@ -51,11 +51,6 @@ define('ECAS_LOGIN_EXLUDE_PAGES', 0);
 define('ECAS_LOGIN_INCLUDE_PAGES', 1);
 
 /**
-  * ECAS group for CEM team users.
-  */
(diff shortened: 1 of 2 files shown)


Fefaine (commented):
what if 'cem' user role does not exist in the site that uses the module?

gillesdeudon:
> what if 'cem' user role does not exist in the site that uses the module?

Good point!. I only focused on the constant issue. The question is what do we do if the cam role does not exist. Do we keep the logic around the "administrator" role assignment or we bypass the full process of automatic assignment?

PR #2976, opened 2021-04-28
NEPT-2922: TMGMT OG - unable to accept translation

piotrsmykaj:
### Description

Fixed access callback for menu default local task item in tmgmt_og.

### Change log

- Fixed: tmgmt_og_menu_alter().

### Checklist

### DIFF
--- a/profiles/common/modules/custom/tmgmt_og/tmgmt_og.module
+++ b/profiles/common/modules/custom/tmgmt_og/tmgmt_og.module
@@ -115,6 +115,24 @@ function tmgmt_og_menu_alter(&$items) {
       $item['title'] = isset($item['title']) ? $item['title'] : '';
       $item['title arguments'] = isset($item['title arguments']) ? $item['title arguments'] : array($item['title']);
       $item['title callback'] = isset($item['title callback']) ? $item['title callback'] : 't';
+
+      // If it's menu default local task and has missing access callback we use
+      // the callback from the parent.
+      if (($item['type'] == MENU_DEFAULT_LOCAL_TASK) && !isset($item['access callback'])) {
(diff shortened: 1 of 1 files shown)


profiles/common/modules/custom/tmgmt_og/tmgmt_og.module
```diff
@@ -115,6 +115,24 @@ function tmgmt_og_menu_alter(&$items) {
...
+          if (isset($parent['access callback'])) {
+            $item['access callback'] = $items[$parent_path]['access callback'];
+          }
+          if (!isset($item['access arguments']) && isset($parent['access arguments'])) {
+            $item['access arguments'] = $items[$parent_path]['access arguments'];
+          }
+        }
+      }
+
       $item['access arguments'] = isset($item['access arguments']) ? $item['access arguments'] : array();
```
Fefaine, line 136: I checked the issue and what you report, and actually I would rather use this solution
https://www.drupal.org/node/553368 

$item['access arguments'] = isset($item['access arguments']) ? $item['access arguments'] : array(0);

piotrsmykaj: @Fefaine 
Please be aware that a few lines below 0 will be replaced to 3 and with that approach, you check user access to "admin" path instead of pass user through 
`entity_access($op, $entity_type, $entity = NULL, $account = NULL)`

Fefaine: Okééééé ! Actually there is another issue when the job is outside of og context, it's currently setting the argument to 6 .. 🤣 
So you are fixing this too, for me it's fine if QA approves.

PR #2979, opened 2021-05-04
NEPT-000: Nexteuropa tracked changes avoid error.

joaocsilva:
### Description

When a content type do not have any wysiwyg field present, the nexteuropa_trackedchanges will throw a error when accessing a node edit form.

This is an example on the Info site for a content type that do not have a wysiwyg field.
Screenshot 2021-05-04 at 09 44 40

The fix that I propose is just to check if variable `CKEDITOR` exists.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

[Insert commands here]

### Checklist

### Notes

Please let me know if I should create a ticket for this.
Thank you

### DIFF
--- a/profiles/common/modules/features/nexteuropa_trackedchanges/js/nexteuropa_trackedchanges_ckeditor.js
+++ b/profiles/common/modules/features/nexteuropa_trackedchanges/js/nexteuropa_trackedchanges_ckeditor.js
@@ -21,18 +21,20 @@
           }
         }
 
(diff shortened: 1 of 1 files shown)


Fefaine:
@joaocsilva  i tried to reproduce this, but it seems unclear
Why would you use the HTLM + tracked changes on a field that does not have any wysiwyg ?
That filter is set per field, so why not just setting the filter to plain text or HTML ?

PR #2985, opened 2021-05-05
NEPT-2926: Only launch the tracked changes if one field is processed.

Fefaine:
### Description

Only launch the tracked changes if one field is processed.

### Change log

- Changed:  Conditions on which to add the js file

### Commands

[Insert commands here]

### Checklist


Fefaine:
@joaocsilva  what do you think?

joaocsilva:
Hi @Fefaine , I tested in my side and it looks great 👍  thanks

PR #3011, opened 2022-01-07
NEPT-2958: PHP 7.4 update parameters order of implode().

julien-:
### Description

PHP 7.4 update parameters order of implode().


Fefaine:
This was already included in this PR 
https://github.com/ec-europa/platform-dev/pull/3006/files

PR #3013, opened 2022-01-11
NEPT-2959: Update 'token' 1.7 => 1.8.

julien-:
### Description

After upgrading to PHP 7.4: 

Warning: strnatcmp() expects parameter 2 to be string, array given in token_asort_tokens() (line 579 of /home/ec2-user/environment/maritimeforum-dev/build/profiles/multisite_drupal_communities/modules/contrib/token/token.module).

### Change log

- Changed: Upgrade token module version


Fefaine:
This is also already done here https://github.com/ec-europa/platform-dev/blob/release-2.6/resources/multisite_drupal_standard.make#L758

PR #3039, opened 2022-03-03
Feature/nept 2985 platform update

Aporie:
### Description

POC of all modules and libraries update.


julien-:

